PR #79, opened 2026-08-16
feat(application): set an existing application's image registry (ankra-tpwy)

CodeStaple:
Bead: `ankra-tpwy` · Linear: PLA-756 · Ankra ticket #1067 (Smartoptics) · Backend: ankraio/cluster#1197

## Why

An application publishes into the organisation's own Ankra registry project unless it declares otherwise. That declaration could only be made **at create time** — there was no update endpoint, no portal field, and no CLI flag. An organisation whose builds push to their own Harbor therefore had no way to correct an already-onboarded application: the readiness lane kept reporting its images as never published while the application ran healthily, and demo mode (which pulls from the resolved registry) stayed closed to them entirely.

## What

Three subcommands over the new `GET`/`PUT /api/v1/org/applications/{id}/image-registry` endpoints landing in ankraio/cluster#1197.

```
ankra application registry get <application-id>
ankra application registry set <application-id> --url oci://<host>/<project> --credential <name>
ankra application registry clear <application-id>
```

**`get`** reports the effective registry: the stored declaration, whether it is one at all, the host and project it resolves to, the primary component's expected image repository, and each component's own — so the coordinates can be compared against where the builds actually push.

**`set`** declares a registry. `--url` is required (scheme optional; `artifact.example.com/commerce` works). `--credential` names the organisation registry credential that authenticates to it — the same credential resource the Helm OCI chart registries already use. Optional `--api-url`, `--pull-secret`, `--username-secret`, `--password-secret`, and `--manage-actions-secrets`.

Setting a registry **without** `--credential` is allowed but warns on stderr: without one Ankra can describe where the images live and neither read nor pull them, so builds keep reporting as never published. The warning goes to stderr so `-o json|yaml` stdout stays parseable.

`--manage-actions-secrets` is off by default and the backend refuses it without a credential name — Ankra never mints robots for a registry it does not administer, and overwriting a customer's push robot would break their build.

**`clear`** returns the application to the organisation's own registry project. It confirms first (declining exits 4), and sends `"image_registry": null` as an **explicit null** — omitting the key would leave the declaration in place.

## Notes

- `internal/client` gains `GetApplicationImageRegistry` / `UpdateApplicationImageRegistry`, mirrored into the `APIClient` interface in `cmd/services.go` and the `baseMock` stub set in `cmd/e2e_test.go`, per the repo invariant.
- `UpdateApplicationImageRegistryRequest.ImageRegistry` deliberately carries **no** `omitempty`: the null is load-bearing.
- Requires the backend endpoints; without them `set`/`clear` surface the platform's 404 with exit code 3.

## Testing

`cmd/application_registry_test.go` covers subcommand registration, JSON rendering, full flag mapping (including URL trimming), the missing-`--url` usage error (exit 2), the missing-credential warning, the explicit-null clear (asserted on the marshalled request), and a declined confirmation (exit 4).

`make test` (`-race`), `make lint` (0 issues), and `make build` all pass. Changelog entry added under `## Unreleased`.

## Follow-up (not in this PR)

Portal still has no field for this — tracked alongside the backend PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,24 @@
 # Ankra CLI Changelog
 
+## Unreleased
+
+### Added
+
+- **`ankra application registry` points an existing application at a container
+  registry you operate.** An application publishes into the organisation's own
+  Ankra registry project unless it declares otherwise, and that declaration
+  could previously only be made when the application was created — so an
+  organisation whose builds push to their own Harbor had no way to correct an
+  already-onboarded application, and Ankra kept reporting its images as never
+  published. `registry get` shows the effective registry, the host and project
+  it resolves to, and the image repository each component is expected to
+  publish to, so you can compare them against where your builds actually push.
+  `registry set --url oci://<host>/<project> --credential <name>` declares one;
+  `registry clear` returns the application to the organisation's own registry.
+  Setting a registry without `--credential` is allowed but warns, because
+  without a credential Ankra can describe where the images live and neither
+  read nor pull them.
+
 ## v0.10.1 — 2026-08-13
 
 ### Fixed
--- a/cmd/application_registry.go
+++ b/cmd/application_registry.go
@@ -1,0 +1,173 @@
+package cmd
+
+import (
+	"errors"
+	"fmt"
+	"strings"
+
+	"github.com/spf13/cobra"
+
+	"ankra/internal/client"
+)
+
+// The application image-registry commands. An application publishes into the
+// organisation's own Ankra registry project by default; one that declares a
+// registry it already operates publishes there instead, and Ankra reads its
+// image tags back with the organisation registry credential the declaration
+// names. Until these commands existed the declaration could only be made at
+// create time, so an already-onboarded application could not be pointed at a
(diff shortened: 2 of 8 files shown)


CodeStaple:
Added a second commit bumping the Go toolchain to 1.26.6.

`govulncheck` was failing this PR for a reason unrelated to it: four Go standard-library advisories published after v0.10.1 was cut are reachable from this code — `crypto/tls` handshake limits (GO-2026-6218, GO-2026-6090), `encoding/asn1` recursion depth (GO-2026-5972), and the `golang.org/x/net/idna` Punycode rejection through `net/http` (GO-2026-5026) — all fixed in 1.26.6. master last ran green on 2026-08-13, before they landed, so the first branch to open a PR since inherits the failure.

Reproduced locally on 1.26.5 (4 vulnerabilities) and confirmed clean on 1.26.6 (`No vulnerabilities found`). The module's `go 1.25.0` language version is untouched; `make test`, `make lint` and `make build` all pass identically on the new toolchain.

Happy to split this into its own PR if you would rather keep it separate from the registry commands.

PR #88, opened 2026-08-17
fix(upcloud): stop sending a colliding default network range; add --cni and --include-dns

CodeStaple:
## The bug

`ankra cluster upcloud create` hardcoded `--network-ip-range 10.0.0.0/16` and always sent it. UpCloud rejects overlapping networks, so on any account that already holds a `10.0.x` private network the create is **accepted**, then wedges at the network step with the reconcile loop failing forever:

```
UpCloud API 409: Network address 10.0.0.0/16 overlaps with an existing private network 10.0.0.0/24.
```

Reproduced live today (cluster `379f68e0`, nine resources failing) while the identical create through the API lane — which omits the field — auto-picked a free range and provisioned first try. The CLI was the only surface pinning a literal default.

## The fix

- `--network-ip-range` defaults to empty; an unset range is **omitted** from the request (`omitempty`), so the platform picks a range that is free in the account. Passing a range still pins it, and the help text now says exactly that.
- Parity gaps closed while in the file: `--cni` (k3s: flannel/calico/cilium; kubeadm always cilium) and `--include-dns` (delegated ankra.cc subdomain opt-out, default on) — completing the `external-cloud-provider` / `include-networking` / `include-dns` trio the API and wizard already expose.

## Test

New client test pins the wire semantics: unset range absent from the body, `cni` and `include_dns` present. `go test ./cmd/... ./internal/client/...` green, gofmt clean on touched files, and the fixed binary verified live: the exact previously-failing command re-run with a locally built binary provisioned a cluster successfully (see follow-up comment with the run).

## Related

The portal wizard prefills the same `10.0.0.0/16` literal and would hit the same 409 — filed separately. A server-side create preflight that rejects an overlapping range at request time (instead of accepting and failing in reconcile) is also filed.

https://claude.ai/code/session_01AVjTVS6VVy8azWdVayMMsT

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,29 @@
 
 ## Unreleased
 
+### Fixed
+
+- **`ankra cluster upcloud create` no longer collides with existing networks.**
+  The command shipped `--network-ip-range` with a literal `10.0.0.0/16` default
+  and always sent it, so on any account that already had a `10.0.x` private
+  network the create wedged at the network step with `UpCloud API 409: Network
+  address 10.0.0.0/16 overlaps with an existing private network`. The flag now
+  defaults to empty and an unset range is omitted from the request, letting the
+  platform pick a range that is free in your account — the same behaviour the
+  portal's API and AI lanes always had. Pass `--network-ip-range` only to pin
+  a specific range.
+
+### Added
+
+- **`ankra cluster upcloud create --cni`** selects the container network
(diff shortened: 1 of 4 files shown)


CodeStaple:
Superseded by #127 (rebased onto master; only the --cni flag remained — the network-range fix and --include-dns were already on master), merged as af6153f23.

PR #131, opened 2026-08-22
release: promote the changelog to v0.13.0 (ankra-jo5qx)

Myothas:
## What & why

Promotes `## Unreleased` to `## v0.13.0 — 2026-08-22` so the tag can be cut. Fifteen commits sit on master since v0.12.0 — among them the four `cluster apply` fixes that make the platform's IaC export round-trip, the stack-profiles lifecycle commands, `--sort`/`--order` on the main list commands, `--include-dns` on every provider create, and the bastion groups for Scaleway/Proxmox/Morpheus — and none are in a tagged release. `release.yml` extracts release notes by matching the `## v<version>` heading, so this promotion is what makes the release carry the written notes instead of auto-generated ones.

Follows the same structure as the v0.12.0 promotion (#106): heading retitled with the date, a short thematic intro paragraph under it, no empty Unreleased section left behind.

The section was audited commit-by-commit against master: every user-visible change since v0.12.0 has an entry; the only commits without one are CI/chore (#113, #120, #121), which get none by convention. No breaking changes → minor bump, straight release without RCs (v0.12.0 precedent).

After merge: tag the squash commit `v0.13.0` and push the tag; `release.yml` builds the six-target matrix and creates the GitHub release. Note: the two darwin matrix builds still run on GitHub-hosted `macos-latest`, so if the org Actions billing issue (ankra-m5m9d) is still live they will not start — the release will be held/reported rather than shipped incomplete.

Bead: ankra-jo5qx

## Test plan

- Changelog-only change; YAML/markdown untouched by code gates. Pre-commit hook (go test + golangci-lint) passed on commit.
- Verified `release.yml`'s note-extraction matches the `## v0.13.0` heading format used by every prior release section.

## Docs checklist

- [x] No user-facing command/flag changes — no docs needed
- [ ] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); ensure `Short`/`Long`/`Example` on new commands are written for docs readers
- [ ] Broader behaviour changes — docs updated in [ankra-docs](https://github.com/ankraio/ankra-docs) (link the PR): 

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #130, opened 2026-08-22
release: cut v0.13.0-rc0

CodeStaple:
Cuts **v0.13.0-rc0** — the first RC of the v0.13.0 minor line, carrying the Smartoptics support-sweep CLI work (bead `ankra-ymkj8`) plus the stack-profile lifecycle and create-flag additions merged this cycle.

CHANGELOG.md only — no source changes. Promotes the `## Unreleased` block to `## v0.13.0-rc0 — 2026-08-22`.

Highlights in this RC:
- `cluster logs --previous` / `-l <selector>` / `--all-containers`, `cluster describe`, `cluster events --for`, `cluster top pods|nodes` (PLA-769)
- `org domain get|set`, `org dns zones`, and `cluster domain` made a non-mutating read (PLA-768/#1076, PLA-770/#1078)
- stack-profiles full lifecycle + drafts options/annotate, `apply --dry-run`
- `--include-dns` on every provider create; `bastion` groups for scaleway/proxmox/morpheus; `upcloud create --cni`
- fixes: `cluster apply` group/values/variables/manifest_base64 round-trip, `--follow=false` bounded read, git-hook repo inspection

Any tag containing `-` publishes as a GitHub prerelease, so this reaches only beta-channel users (`ankra config beta enable`). Tag is cut after merge.

PR #132, opened 2026-08-22
release: revert the premature v0.13.0 changelog promotion (#131)

Myothas:
## What & why

Reverts #131 (`f7a2f8f`). #131 and #130 raced: #130 promoted `## Unreleased` to `## v0.13.0-rc0` for the Smartoptics support-sweep RC and merged 77 seconds before #131, which had been cut from a pre-rc0 base and promoted the same content to a straight `## v0.13.0`. The textual merge was clean, so master ended up with an entry-less `## v0.13.0` section stacked above the real `## v0.13.0-rc0` one — wrong for every path forward: tagging v0.13.0 against it would ship release notes with no entries, and the stable cut after rc0 soak needs to write that section itself with the entries folded in (the v0.11.0-line convention: stable section carries the full list, rc sections retained beneath).

The v0.13.0-rc0 lane is untouched — its tag points below #131's commit, so its notes extracted cleanly and the release built normally.

After this revert, master's changelog is: empty `## Unreleased` (room for the next cycle), then `## v0.13.0-rc0 — 2026-08-22`, then history. Stable v0.13.0 will be promoted properly once the RC has soaked (tracked in ankra-jo5qx — mentioned as context; this PR does not deliver that bead).

## Test plan

Changelog-only revert; pre-commit hook (go test + golangci-lint) passed on commit. Verified the reverted file's top-of-file state matches the post-rc0-promote convention.

## Docs checklist

- [x] No user-facing command/flag changes — no docs needed
- [ ] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); ensure `Short`/`Long`/`Example` on new commands are written for docs readers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,15 +1,6 @@
 # Ankra CLI Changelog
 
-## v0.13.0 — 2026-08-22
-
-Makes the platform's IaC export round-trip. `cluster apply` now carries an
-exported ImportCluster back without dropping addon values, stack variables,
-group labels, or base64-encoded manifests — the clone-edit-apply loop deploys
-what the file says, or fails naming what it could not read, instead of
-silently installing chart defaults. Stack profiles gain the portal's whole
-lifecycle rather than just the read half, and the main list commands learn
-`--sort`/`--order` so `cluster list --sort created --order desc` puts the
-newest clusters first.
+## Unreleased
 
 ## v0.13.0-rc0 — 2026-08-22
 


Myothas:
Re the [low] finding: the premise isn't quite right — the changelog still fully tracks that work. The itemized entries for the IaC round-trip fixes, the stack-profiles lifecycle, and `--sort`/`--order` were never part of #131's diff; they live in the `## v0.13.0-rc0 — 2026-08-22` section directly below (promoted by #130), which is the section the rc0 release notes were extracted from. The only thing this revert deletes is #131's 10-line stable-release intro paragraph, and that prose summarizes an already-RC-released set, so `## Unreleased` (not-yet-released changes) would be the wrong home for it. It gets rewritten as part of the real `## v0.13.0` stable promotion after the RC soaks — per the v0.11.0-line convention where the stable section carries the full merged notes above the retained rc sections — and is trivially recoverable from #131 if wanted verbatim.

PR #128, opened 2026-08-22
feat(org): manage MCP tool servers from the CLI (ankra-4mr54)

Myothas:
## What

`ankra org mcp-servers` - a full command group for the org MCP tool-server registry, over the bearer-PAT lane cluster#1646 added (`/api/v1/org/mcp-servers` + `/api/v1/org/secret-slots`):

```
ankra org mcp-servers catalog                      # curated adapters (Sentry, PostHog, GitHub, Grafana, OneUptime)
ankra org mcp-servers list | get <name-or-id>
ankra org mcp-servers add sentry --url https://mcp.sentry.dev/mcp --adapter sentry \
    --secret-header Authorization                  # prompts hidden; value goes to a secret slot
ankra org mcp-servers update|remove|enable|disable <name-or-id>
ankra org mcp-servers health|tools <name-or-id>
ankra org mcp-servers grants|grant|revoke-grant <name-or-id> [tool]
```

## Highlights

- **Secrets never travel in plaintext**: `--secret-header Key[=Value]` (hidden prompt when no value) creates a Vault-backed secret slot labeled `"<server> <Key>"` and registers the server with the returned `${SECRET_SLOT:<id>}` sentinel - matching the backend's refusal of plaintext under sensitive header names. The tests assert the raw value never appears in the register payload or output.
- **Adapter-guided setup**: with `--adapter` and no `--allowed-tools`, the allow-list seeds from the catalog's `recommended_tools` for that adapter.
- Name-or-UUID resolution with an "Available servers: ..." hint on miss (exitNotFound); `remove` confirms unless `--yes`; structured `-o json|yaml` on the read commands; pydantic 422 details flattened to readable `field: msg` lines; RBAC denials surface as the standard permission error (exit 7).
- Three-place invariant kept: client methods + `APIClient` interface + `baseMock` stubs; README org row + CHANGELOG Unreleased entry updated; docs-quality Short/Long/Example on every command (the CLI reference regenerates from them).

## Tests

`go test ./...` green (cmd 2.7s, internal/client 10.1s): httptest client coverage (catalog/list/create body incl. sentinel env, 409/422/404 mapping, null-clear update, enable/disable, grant/revoke, secret slots) and command tests via the mock recipe (table render, add composition + seeding, secret-not-echoed, grant, remove --yes, resolve-miss). golangci-lint: 0 issues.

Bead: ankra-4mr54

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #133, opened 2026-08-22
fix(org): mcp-servers review follow-ups - orphaned slots, piped secrets, changelog splice (ankra-4mr54)

Myothas:
## What

Addresses all six findings from #128's Ankra AI review (merged before the review was read - the ledger is on that PR):

- **CHANGELOG splice** [medium]: restores the `ankra cluster logs --previous` headline the new entry had replaced.
- **Orphaned secret slots** [medium]: `add` now tracks the slots it creates and best-effort deletes them on any later failure (fresh 30s context so an expired deadline can't block cleanup); the error names the cleanup, undeletable slot ids go to stderr.
- **Piped stdin** [medium]: one shared `bufio.Reader` per invocation - two piped `--secret-header Key` values both land instead of the second seeing EOF.
- **Help text** [low]: steers to the hidden-prompt/stdin path; warns inline `Key=Value` lands in shell history and `ps`.
- **204 empty body** [low]: `decodeMCPResponse` skips decoding empty bodies at all 13 sites.
- **`--url` for stdio** [low]: stays required (the backend contract requires url for every transport; stdio rows are stored, never dialed) - the help now says so and suggests a `cmd://<binary>` placeholder.

## Tests

Five new tests (orphan cleanup, undeletable-slot warning, two piped secret headers, two 204 paths) plus a flag-state reset helper for add-invoking tests. `go test ./...` green; golangci-lint 0 issues.

Bead: ankra-4mr54

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@
   configuration is also the laziest one to type. Servers resolve by name or
   id everywhere, deletion confirms unless `--yes`, and every read supports
   `-o json|yaml`.
+- **`ankra cluster logs --previous` reads the log a crash-looping container
   left behind.** When a container is in CrashLoopBackOff the only output
   worth reading belongs to the instance that already died, and nothing in
   the CLI could reach it — the one case where you most want to avoid handing
--- a/cmd/mcp_servers.go
+++ b/cmd/mcp_servers.go
@@ -186,15 +186,22 @@ non-secret values only; the backend refuses plaintext under
 sensitive-looking names). --secret-header stores the value in an
 organisation secret slot labeled "<server-name> <Header>" and registers the
 server with the slot's "${SECRET_SLOT:<id>}" sentinel, so the secret never
-lands in the server record. Pass the value inline (--secret-header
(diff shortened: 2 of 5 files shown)


Myothas:
Finding ledger for the changes-requested review (its footer still names 3a573b6 - no re-review posted for d47e9a6 within ~14 minutes, noting per protocol rather than waiting forever). All three findings are FIXED in d47e9a6, none argued away:

- [medium] usage-error bypassing slot cleanup: closed by construction - every --secret-header pair is validated in a pre-pass BEFORE any slot is created, so the usage error fires while zero slots exist; the in-loop failure paths still route through cleanupSlots. Test pins zero slots on a malformed second pair.
- [low] duplicate secret-header keys: refused up front (also a key colliding with a plaintext --header), before any slot exists. Tests pin both refusals.
- [low] empty-body success on reads: decodeMCPResponse takes an explicit allowEmpty; only the four acknowledgement calls pass true, the nine reads fail loudly on an empty 200 ("the server returned an empty body"). Tests pin both sides of the boundary.

CI green on d47e9a6; golangci-lint 0 issues.

PR #134, opened 2026-08-22
fix(domain): report the held removal, split the blockers you cannot clear, and say what a root switch touches (ankra-ymkj8.16)

CodeStaple:
Bead: ankra-ymkj8.16
Linear: PLA-771 (Ankra ticket #1079, Smartoptics)
Backend: ankraio/cluster#1687

The CLI half of making the root-domain switch converge. Follow-up to
ankra-cli#122/#123/#124 (ankra-ymkj8.2).

## `ankra cluster domain <cluster> --remove` reports the hold

The customer removed a cluster's domain, watched it read `none`, and watched it
read `active` again fifty seconds later under the same label. Backend-side the
removal now holds (cluster#1687); CLI-side the read has to *say so*, because
state `none` alone is exactly the ambiguity they were stuck in — a domain that
is gone and one that is between reconciler passes look identical.

```
Domain: (none)
State:  none
Opted out: yes

This cluster's domain was removed and the removal is held: nothing will re-create it.
Run 'ankra cluster domain so-development --enable' when you want it back.
```

## A refused `ankra org domain set` separates blockers you can clear from ones you cannot

```
Cluster domains still under the current root:
  so-development  axmndle4sl.qh4thi04cs.ankra.cc  (active)
  Remove each with: ankra cluster domain <cluster> --remove
  Each of these clusters runs Ankra's external-dns against its own zone.
  The removal is held once it completes, so external-dns cannot bring the
  zone back; 'ankra cluster domain <cluster> --enable' is what restores it.
  Re-enabling keeps the zone's label, and with it external-dns's
  --txt-owner-id and your GitOps paths - a root switch never re-labels.

DNS records still under the current root:
  CNAME  chat.qh4thi04cs.ankra.cc  (active)
  Remove each with: ankra org dns delete <record>

DNS records Ankra owns and re-creates:
  A  *.smdjc5s3hx  (created by playground_provisioner)
  Deleting these does not clear them - the lane that owns each one writes
  it again on its next pass. Remove what publishes them instead.

Playground environments publishing those records:
  playground  5f2a1c3e-...  (ready)
  Destroy each with: ankra cluster playground destroy <cluster_id>
  The switch cannot be attempted while a playground exists.
```

The middle two sections used to be one, which is how the reporter ended up
deleting `*.smdjc5s3hx` and watching `playground_provisioner` write it back two
minutes later. Records arriving without `created_by` — an older backend — are
treated as the admin's, so nothing that was deletable before starts reading as
un-deletable.

The cluster-zone paragraph answers ask 1: what those clusters' external-dns will
do about the removal, and that labels come from the cluster id so a switch never
re-labels.

## `--help` answers the question that decides whether the switch is safe

Their words: *"whether adopting it as the Ankra root leaves them alone is the
single fact that decides whether this is safe, and it is not written down
anywhere."* It is now, in both `ankra org domain --help` and `set --help`:

> Ankra creates one subzone in the domain you register — `<org_short_id>.<domain>`
> — and works only inside it. It does not read, write or delete records anywhere
> else in the domain. Records you already publish at the apex or under any other
> name are untouched by registering the domain, by the switch itself, and by
> everything Ankra does afterwards.

Plus a **WHAT A SWITCH DOES NOT CHANGE** section on label and `--txt-owner-id`
stability. Both are read off `EnsureOrganisationZone` and `ClusterDNSZoneLabel`.

## `ankra cluster playground destroy <cluster_id>`

The refusal now points at a playground as the thing to remove, and there was no
CLI verb to remove it — `DELETE /api/v1/org/clusters/playground/{id}` has existed
since playgrounds shipped, but `create` and `status` were the only commands
wired to it. Idempotent, like the endpoint.

## Tests

- `TestRunClusterDomain_ReportsAHeldRemoval`
- `TestRunOrgDomainSet_SeparatesPlatformOwnedRecordsAndNamesThePlayground` —
  including that neither kind of record leaks into the other's section
- `TestRunOrgDomainSet_NamesTheExternalDNSInteractionOnBlockingZones`
- `TestOrganisationDomainBlockedMessage_TreatsAnUnknownWriterAsTheAdmins` — the
  older-backend case

`TestRunClusterDomain_RemoveDoesNotEmitTheEnableHint` asserted on the substring
`--enable`; it now asserts on the empty-cluster hint it was written to guard
("this cluster has no public domain, create one"), and additionally that the
removal path states the hold. Naming `--enable` as the way to withdraw a hold is
a different statement, and the removal path should make it.

`go build ./...` and `go test ./...` green.

https://claude.ai/code/session_011DgQPJJkGkGjRYAFJ56iJ4

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,50 @@
 
 ## Unreleased
 
+### Added
+
+- **`ankra cluster playground destroy <cluster_id>` tears the organisation's
+  playground down.** The platform has served the DELETE since playgrounds
+  shipped; only the CLI could not reach it, so `create` and `status` had no
+  matching way out. It is idempotent — an environment already deprovisioning
+  or removed answers its current phase rather than an error. It also has a
+  second job: a live playground publishes a wildcard DNS record in the
+  organisation's zone, that record is reconciled rather than written once,
+  and it is therefore the one blocker of `ankra org domain set` that deleting
+  cannot clear. Destroying the environment is what clears it.
+
+### Changed
+
+- **`ankra cluster domain <cluster> --remove` now holds.** Removing a
+  cluster's domain and watching it reappear about fifty seconds later, active
+  and under the same label, was the reason the documented root-domain switch
+  could not be completed: every removal was reverted before the next one
+  could be made. The removal is now recorded as a deliberate act, so nothing
+  re-creates the zone — not the external-dns Ankra runs on the cluster, and
+  not the discovery that mints zones for clusters that have none. A read
+  reports `Opted out: yes` for as long as the hold stands, which is what
+  distinguishes a domain that is gone from one that is between passes.
+  `--enable` withdraws the hold and re-creates the zone under the
+  organisation's current root, under exactly the name it had before.
+- **A refused `ankra org domain set` now separates the blockers you can clear
+  from the ones you cannot, and names what to remove instead.** DNS records
+  the platform publishes and re-asserts are listed apart from your own, with
+  the writer named, because telling an admin to `ankra org dns delete` a
+  record the playground provisioner writes back on its next pass is advice
+  that cannot work. Live playground environments are listed outright, with
+  the command that destroys them. The cluster-domain section now also says
+  what those clusters' external-dns will do about the removal, and that a
+  root switch never re-labels a zone — labels come from the cluster id, so
+  `--txt-owner-id` and any GitOps path built from one survive the switch
+  unchanged.
+- **`ankra org domain --help` and `ankra org domain set --help` state what
+  Ankra does to the domain you register.** Ankra creates one subzone,
+  `<org_short_id>.<domain>`, and works only inside it: records already
+  published at the apex or under any other name are never read, written or
+  deleted, by the switch or by anything Ankra runs afterwards. A domain that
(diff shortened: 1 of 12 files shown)


CodeStaple:
Both review findings addressed in 69630e4.

**`destroy` wrote with `fmt.Printf`, untested** — right, and the two were the
same problem: `Printf` writes past the writer a test sets with `SetOut`, so the
verb could not be asserted on. It now writes through `cmd.OutOrStdout()` like
`cluster_domain` and `org_domain` do, with `TestPlaygroundDestroyPrintsTheClusterIDAndPhase`
(cluster id **and** phase — the phase is what says teardown was scheduled rather
than finished) and `TestPlaygroundDestroySurfacesTheServerError`.

`create` and `status` keep their `Printf`. They are the existing shape of this
file and not what this PR changes; converting them would be a drive-by.

**Section assertions panic instead of failing cleanly** — correct, and worse
than it looks: the checks above that line are `t.Errorf`, so a regression that
dropped a section header would reach the slice and panic with an out-of-range
error rather than reporting *which* header went missing. Both offsets are now
checked for `>= 0` and for their order, with `t.Fatalf` carrying both values.

`go build ./...` and `go test ./...` green.

PR #135, opened 2026-08-23
feat(application): reach the 13 bearer-lane operations the CLI could not (ankra-hcgru)

Myothas:
## What

The 13 bearer-lane application operations the CLI could not reach.

```
ankra application env-secrets   list | set | delete | apply
ankra application auto-deploy   get  | set
ankra application settings      get  | set
ankra application manifest-addon get | diff | install | unpublish | delete
ankra application demo fix-build
```

## Why

Diffing the cluster router's `/api/v1/org/applications` lane against `internal/client`: **58 operations are mounted, 6 are 410 tombstones** (deploy preview/execute/runs, package-visibility, make-public, all served by `servePackageVisibilityGone` / `deployGoneTokenHandler`), and of the **53 live ones the CLI reached 39**. No CLI call targeted a route that does not exist, so this is purely the missing 14. (53 rather than 52 because cluster#1717 adds the fix-build bearer twin; with `demo fix-build` here, the CLI now reaches **every live operation of the lane**.)

`env-secrets` is the one that closed a broken journey rather than adding a convenience. The keys come from the application's own generated manifests, so a pipeline could create an application, build it, deploy it, and then stop at the one step that made it run. It is the platform answer to the `kubectl create secret generic <app>-env --from-literal=...` line the setup pull request used to end on, and it was portal-only.

The manifest-addon five are the tail of a lane the CLI already half-served: `publish-addon` and `published-addon` shipped, so a user could publish an application's manifests into the add-on catalog and then not inspect, diff, install, unpublish or delete what they published.

## Contract details a naive client gets wrong

- **The env-secret value is not trimmed.** `--value` is taken verbatim and a piped value loses exactly the one line ending the pipe added. The CLI's existing `resolveSecretInput` trims every path, which is right for an API key and wrong for an application's environment: a PEM block, base64 with padding, or a password ending in a space would be stored as a value the workload never matches, and nothing would show it afterwards because **no route hands a stored value back**. `resolveEnvSecretValue` says this at its definition rather than leaving it to be rediscovered. The guard is `TestApplicationEnvSecretSetPreservesSignificantWhitespace`; reintroducing the trim fails two of its four cases (verified by injecting it).
- **Clearing sends an explicit null.** `PUT /settings` requires `ci_runner_label` to be present, so an omitted key is a 422 missing-field error rather than a clear. `UpdateApplicationSettingsRequest` therefore has no `omitempty`, matching what `UpdateApplicationImageRegistryRequest` already does for the same reason.
- **`auto-deploy` sends off as an explicit false**, and `--enabled` must be given explicitly. Turning unattended deployment on and turning it off are both deliberate acts and neither is a safe default to infer from a bare `set`.
- **`--input key=value` splits on the first `=` only**, so a value containing one (a DSN, a query string) survives, and an `--input` with no `=` is a usage error rather than a key silently answered with `""`.

## Safety

- Every destructive command confirms and exits 4 when declined: `env-secrets delete`, `manifest-addon unpublish`, `manifest-addon delete`.
- `unpublish` and `delete` are different acts and the prompts say which is which: unpublishing withdraws the catalog entry and leaves installations running, deleting undeploys every installation that came from it.
- The secret value never reaches stdout. `TestApplicationEnvSecretSetNeverEchoesTheValue` pins it, so a `set` inside a script with `-o json` cannot write the secret into whatever captured the output. The "storing is not applying" hint goes to stderr for the same reason.
- Secret keys are path-escaped (`TestSetApplicationEnvSecretEscapesTheKey`).

## Verification

- `go build ./...`, `go test -race -count=1 ./...` (whole suite green), `make lint` → **0 issues**.
- 14 new client tests assert the method, path, query and body of every new call, because a path typo is a runtime 404 that every command-wiring test passes.
- 21 new command tests cover registration, both auto-deploy states, clear-versus-label, the input parser, the whitespace contract, and each declined confirmation.
- No em dashes in the new copy; the six pre-existing `gofmt -l` hits in this repo are untouched files.

## Related

- **`demo fix-build` needs [cluster#1717](https://github.com/ankraio/cluster/pull/1717) deployed** (`ankra-961wq`), which adds the bearer twin it calls. cluster deploys from main continuously and the CLI releases on tags, so #1717 will be live well before this reaches a released binary; merge it first regardless. `demo build` is the check that reports the missing image and it was already in the CLI, so the check and its remedy are finally on the same surface.
- The other 12 session-only application routes are documented policy in `rollback.go`, `codereview.go` and `handlers_ai.go`, and are deliberately not exposed here.
- `ankra-orhek` tracks the same audit's MCP finding.

Bead: ankra-hcgru


Myothas:
All three fixed in `aea3546`. They were one class of defect: an input that is almost certainly a mistake, accepted silently.

**[medium] `--value ""` bypassed the empty guard.** Right, and the inconsistency was the tell: stdin and the prompt both refused empty while the flag did not. All three paths refuse now, and the refusal points at `env-secrets delete` for the case where clearing was what was meant. `TestApplicationEnvSecretSetRefusesAnExplicitlyEmptyValue` covers both `--value=` and `--value ""`.

**[low] `PathEscape` leaves dot segments.** Confirmed and closed. The key is now validated to the platform's own rule before it reaches a path: `validateEnvSecretKey` mirrors `manifestenv.IsValidKey` (letter or underscore, then letters/digits/underscores, 1 to 63 characters), which is what the backend enforces anyway, so this can only turn a 400 into a usage error - and it does so *before* the path is built. The client keeps escaping the key as defence in depth. `TestApplicationEnvSecretRefusesAKeyThatIsNotAnEnvironmentVariableName` covers `..`, `.`, `A/B`, a leading digit, spaces, dashes and empty across both `set` and `delete`, and asserts an ordinary key still works.

Worth saying plainly that the consequence you traced is the reason this was worth fixing rather than arguing about: `DELETE .../env-secrets/..` normalising onto the application resource is the delete-application route.

**[low] Duplicate `--input` keys.** Agreed, and for the reason you gave: there is one value per input, so last-one-wins can never be what someone meant, and the usual cause is a variable expanded twice. It is a usage error naming the duplicated key. `TestManifestAddonInstallRejectsADuplicateInputKey`.

`go test -race -count=1 ./...` green, `make lint` 0 issues.

Myothas:
Both fixed in `246125ed`.

**Hint in structured mode.** Gated on the resolved output format. `TestApplicationEnvSecretSetHintIsHumanOutputOnly` asserts the human run keeps the hint and the `-o json` run does not.

**Helper name.** Kept the reuse and documented it where the question comes up, rather than renaming across ~40 call sites for a diff nobody can review. The point is that `applicationResourceRequest` builds no path of its own: it is the bearer JSON request path for this whole surface (auth header, FastAPI `detail` surfacing, status preserved for exit-code classification), and the manifest add-on file passes it a catalog path. Reusing it is what keeps one error contract rather than a second drifting alongside it, which is now stated in `manifest_addons.go`'s package comment.

`go test -race -count=1 ./...` green, `make lint` 0 issues.

PR #137, opened 2026-08-23
feat(org): reach the preview settings from the CLI, and stop 'org domain' reading as the only custom domain (ankra-ymkj8.18, ankra-ymkj8.20)

CodeStaple:
Bead: ankra-ymkj8.18, ankra-ymkj8.20
Reported as: PLA-773 (Smartoptics, #1081)

Two gaps the customer hit on the same day, both from the same root: the organisation's AI-environment settings live on one endpoint, and the CLI wrapped exactly one of its fields.

## 1. The preview settings were unreachable from the CLI (ankra-ymkj8.18)

`ankra org domain get|set` sends only `dns_root_domain`. `demo_base_domain`, `demo_ingress_class_name` and `demo_tls_secret_name` were portal/API-only — and they are precisely the fields you script when standing up on-demand environments on your own domain. The customer got there with a hand-written `PUT`.

```
ankra org ai-environment get
ankra org ai-environment set --demo-base-domain previews.example.com
ankra org ai-environment set --demo-cert-issuer letsencrypt-prod
ankra org ai-environment set --demo-tls-secret ""      # clears that field alone
```

Only the flags you pass are written. The endpoint reads **presence**, not emptiness, so setting one field never clears its neighbours, and a flag passed with an empty value sends an explicit `null`. A `set` with no flags is a usage error rather than an empty `PUT`.

`get` names what an empty field falls back to, and surfaces the backend's own verdict on whether previews will be served over plain http. That verdict is not guessable from the fields: once ankraio/cluster#1737 lands, a preview domain requests a per-preview certificate by default, and whether it can depends on what the staging cluster carries.

```
Preview domain:    previews.smartoptics.dev
Ingress class:     (cluster default)
TLS secret:        (cluster default)
Certificate issuer: (cluster default)

Demos on previews.smartoptics.dev will be served over plain http: the staging
cluster carries no ACME HTTP-01 ClusterIssuer to request a certificate from.
```

`set` prints the same warning, so a write that disables demo TLS says so rather than reporting success and nothing else. The customer's words: *"We would have set this and shipped demos on plain HTTP without ever being told."*

## 2. `org domain --help` read as *the* custom domain (ankra-ymkj8.20)

Reported verbatim: *"'ankra org domain set --help' says the custom domain lives at 'AI > Settings > Workspaces'. Given demo_base_domain and dns_root_domain are different fields with very different blast radius, that help text sending people to one page for 'the custom domain' is how we spent a day on the wrong one."*

Confirmed — and #134 did not fix it: the long help never mentioned "Preview domain" at all. It now names the second field, contrasts the blast radius (one ungated field write vs. a re-root gated by the cluster-zone guard), and sends that reader to `ankra org ai-environment set --demo-base-domain`.

## Verification

- `go build ./...`, `go test ./...` — all pass
- `golangci-lint run ./cmd/... ./internal/client/...` (v2.12.2, the CI pin) — 0 issues
- 6 new tests: the fallback line, the plain-http warning, the warning withdrawing once an issuer is named, present-vs-absent on `set`, empty-value-clears, and no-flags-is-a-usage-error
- Both help texts rendered from a built binary and read back

### DIFF
--- a/cmd/e2e_test.go
+++ b/cmd/e2e_test.go
@@ -595,6 +595,15 @@ func (m baseMock) SetOrganisationDomain(ctx context.Context, rootDomain string) 
 	return nil, errors.New("not implemented")
 }
 
+func (m baseMock) GetOrganisationPreviewSettings(ctx context.Context) (*client.OrganisationPreviewSettings, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (m baseMock) UpdateOrganisationPreviewSettings(ctx context.Context,
+	changes map[string]*string) (*client.OrganisationPreviewSettings, error) {
+	return nil, errors.New("not implemented")
+}
+
 func (m baseMock) MCPCatalog(ctx context.Context) (*client.MCPCatalogResult, error) {
 	return nil, errors.New("not implemented")
 }
--- a/cmd/org_ai_environment.go
+++ b/cmd/org_ai_environment.go
@@ -166,12 +166,15 @@ func renderOrganisationPreviewSettings(cmd *cobra.Command,
 	out := cmd.OutOrStdout()
 	if settings.DemoBaseDomain == "" {
 		_, _ = fmt.Fprintln(out, "Preview domain:    (none - demos use the staging cluster's Ankra subzone)")
-		return nil
+	} else {
+		_, _ = fmt.Fprintf(out, "Preview domain:    %s\n", settings.DemoBaseDomain)
+		_, _ = fmt.Fprintf(out, "Ingress class:     %s\n", orEmptyDefault(settings.DemoIngressClassName))
+		_, _ = fmt.Fprintf(out, "TLS secret:        %s\n", orEmptyDefault(settings.DemoTLSSecretName))
+		_, _ = fmt.Fprintf(out, "Certificate issuer:%s\n", " "+orEmptyDefault(settings.DemoCertIssuerName))
 	}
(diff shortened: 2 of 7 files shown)


CodeStaple:
Both addressed in b610b6c, though the first is not a live bug today.

**[medium] Early return swallows the warning.** Not reachable right now: the backend scopes the warning to the custom-domain path — `demoPreviewTLSWarning` returns nil immediately when `demo_base_domain` is empty, deliberately, since the subzone path is Ankra's own to get right rather than the organisation's to be warned about. So there is no warning to swallow in that branch.

You are right about the coupling, though. Tying the display to a field puts this command back in the business of second-guessing a verdict it explicitly delegates, and it would fail silently the day the backend widens the warning. The renderer now prints `PreviewTLSWarning` unconditionally, after either branch, with `TestRunOrgAIEnvironmentGet_PrintsAWarningEvenWithNoPreviewDomain` pinning it.

**[low] `doOrganisationDomainRequest`.** The endpoint is genuinely shared — the helper has always built `c.BaseURL + "/api/v1/org/ai-environment"`, so nothing was mis-targeted and no response was mis-decoded. But you read it as a copy-paste error, which is exactly the failure the name invites, so it is renamed `doAIEnvironmentRequest` and its doc comment now says why the name is neutral: that one route backs both the root-domain setting and the preview settings, and it does not belong to the domain lane.

PR #138, opened 2026-08-23
fix(org): yaml output keys match json, and stored preview overrides stay visible (ankra-ymkj8.18)

CodeStaple:
Bead: ankra-ymkj8.18
Follow-up to ankraio/ankra-cli#137 — inline review findings that were not addressed before it merged.

## 1. `-o yaml` emitted different keys from `-o json` (user-facing)

`OrganisationPreviewSettings` carried `json` tags only, and `renderOrganisationPreviewSettings` also encodes it with `gopkg.in/yaml.v3`, which ignores `json` tags and lowercases the Go field names. Reproduced against production before the fix:

```
$ ankra org ai-environment get -o yaml     $ ankra org ai-environment get -o json
demobasedomain: ""                         "demo_base_domain": ""
demoingressclassname: ""                   "demo_ingress_class_name": ""
demotlssecretname: ""                      "demo_tls_secret_name": ""
democertissuername: ""                     "demo_cert_issuer_name": ""
previewtlswarning: ""                      "demo_preview_tls_warning": ""
```

Anything scripted against the YAML form was reading keys that do not exist in the documented contract. `yaml` tags added; verified against production that both forms now agree.

## 2. `get` hid stored overrides whenever no preview domain was set

The early return meant a stored ingress class, TLS secret or certificate issuer became invisible — which is exactly when you need to see them. Clearing only `--demo-base-domain` left values on the organisation that nothing could show and that come back into force the moment a domain is set again. They are now listed, with a line saying they are inert until a domain is set. Nothing extra is printed when nothing is stored.

## 3. Flag names were duplicated between `RunE` and `init()`

Two literal lists of the same four flag names. Renaming one side would have left a flag cobra parses, `Changed` never matches, and the user gets "pass at least one of" despite passing a valid flag. Both sites now read one `previewSettingFlagFields` map, and registration panics on a flag with no wire field.

## Verification

- `go build ./...`, `go test ./...` — all pass
- `golangci-lint run ./cmd/... ./internal/client/...` (v2.12.2, the CI pin) — 0 issues
- 4 new tests: YAML keys match JSON and never fall back to the lowercased names; stored overrides visible and marked inert with no domain; nothing extra printed when nothing is stored; every wire field has a registered flag and vice versa
- The YAML fix confirmed against the live production API, not just in tests

### DIFF
--- a/cmd/org_ai_environment.go
+++ b/cmd/org_ai_environment.go
@@ -205,16 +205,30 @@ var previewSettingFlagFields = map[string]string{
 	"demo-cert-issuer":   "demo_cert_issuer_name",
 }
 
+// previewSettingFlagUsage is the help string for each flag in
+// previewSettingFlagFields. Kept beside it so registration can walk the field
+// map itself: a field added without a usage string then fails at init rather
+// than registering no flag at all, which the update loop would skip in
+// silence.
+var previewSettingFlagUsage = map[string]string{
+	"demo-base-domain":   "Wildcard zone demos are published under; empty clears it",
+	"demo-ingress-class": "Ingress class for demo ingresses; empty clears it",
+	"demo-tls-secret":    "Secret holding a certificate for the preview domain; empty clears it",
+	"demo-cert-issuer":   "cert-manager ClusterIssuer to ask instead of the cluster's own; empty clears it",
+}
+
 func init() {
 	registerStructuredOutputFlags(orgAIEnvironmentGetCmd, orgAIEnvironmentSetCmd)
-	for flagName, usage := range map[string]string{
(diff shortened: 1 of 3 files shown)


CodeStaple:
Good catch, and you're right that the map's own comment promised more than the code delivered. Fixed in b255714.

Registration now walks `previewSettingFlagFields` and looks up the usage string, panicking when it is missing — so a wire field added without a usage entry fails at init instead of registering no flag and being skipped in silence by the update loop. A length check covers the other direction (a usage string naming no wire field). `TestPreviewSettingFlagMapsAgree` pins both, so the test is now a backstop rather than the only thing catching it.

PR #143, opened 2026-08-24
docs(org): 'ai-environment --help' says publishing the preview records is yours (PLA-773)

CodeStaple:
Reported as: PLA-773 (Smartoptics, #1081) — end-to-end field trace, 2026-08-24

Docs-only. One surface, `ankra org ai-environment --help`.

## Why this is small

I started this from the customer's trace and initially opened three PRs (docs, portal, CLI help). Two were redundant and are closed: `ankra-docs` master already carries the full account, and portal main already renders `demo_preview_dns_warning` from the server at exactly the spot I was adding a static paragraph. The runtime warning is the better answer and it has shipped.

This is the one surface that had not caught up. The long help explained *where* demos are published on a preview domain and never said *who* makes those hostnames resolve — which is the omission the customer actually complained about:

> That leaves demo_base_domain depending entirely on the customer's own external-dns being healthy, while the ankra.cc path does not, because there you create the zone and records yourself. **Nothing says so.**

The runtime warning covers `get`'s output. It does not cover the help you read *before* you decide to set the field.

## The wording

Taken from the docs page rather than written fresh, so the two do not give slightly different accounts of the same mechanism. It keeps the part I did not have myself: the external-dns credential Ankra provisions for a cluster is scoped to that cluster's own Ankra subzone and nothing else, which is *why* nothing in the platform can publish on a customer's domain — a stronger and more accurate statement than "Ankra does not write there".

It also carries the consequence the customer hit: without the record a demo still deploys and still reports ready, and its URL does not load, and the certificate goes the same way because an HTTP-01 challenge is answered over the hostname being certified.

## Verification

- `go build ./...`, `go test ./...` — pass
- `golangci-lint run ./cmd/...` — 0 issues
- Help rendered from a built binary and read back

### DIFF
--- a/cmd/org_ai_environment.go
+++ b/cmd/org_ai_environment.go
@@ -202,6 +202,12 @@ func renderOrganisationPreviewSettings(cmd *cobra.Command,
 	// Printed whatever the fields say. The backend decides when previews
 	// lack TLS, and tying the display to a field here would put this command
 	// back in the business of second-guessing that.
+	// DNS first: an unresolvable hostname is why the TLS one is usually
+	// there too, and reading the certificate complaint before the reason for
+	// it sends people to the wrong setting.
+	if settings.PreviewDNSWarning != "" {
+		_, _ = fmt.Fprintf(out, "\n%s\n", settings.PreviewDNSWarning)
+	}
 	if settings.PreviewTLSWarning != "" {
 		_, _ = fmt.Fprintf(out, "\n%s\n", settings.PreviewTLSWarning)
 	}
--- a/cmd/org_ai_environment.go
+++ b/cmd/org_ai_environment.go
@@ -56,9 +56,10 @@ over the hostname being certified, so if the name does not resolve Let's
 Encrypt cannot reach the solver and none is ever issued. An unpublished
 preview domain costs you the URL and the TLS together.
 
-'get' says so when the domain answers nothing, and names the wildcard to
-create and the address to point it at. On the Ankra subzone none of this
-applies: the platform provisions the zone and the in-cluster external-dns
+'get' relays the platform's verdict, which names the wildcard to create and
+the address to point it at when the domain answers nothing. A platform that
+reports no verdict at all is called out as that, rather than left to read as
+an all-clear. On the Ankra subzone none of this applies: the platform provisions the zone and the in-cluster external-dns
 publishes each preview record itself.
 
 TLS ON YOUR OWN PREVIEW DOMAIN
--- a/cmd/org_ai_environment.go
+++ b/cmd/org_ai_environment.go
@@ -208,6 +208,14 @@ func renderOrganisationPreviewSettings(cmd *cobra.Command,
 	if settings.PreviewDNSWarning != "" {
 		_, _ = fmt.Fprintf(out, "\n%s\n", settings.PreviewDNSWarning)
 	}
+	// Saying nothing here would be read as "your hostnames resolve", which is
+	// the one thing this platform has not told us.
+	if !settings.PreviewDNSReported && settings.DemoBaseDomain != "" {
+		_, _ = fmt.Fprintln(out,
+			"\nThis platform does not report whether preview hostnames resolve, so the silence above\n"+
+				"is not an all-clear. Check that a wildcard for this domain points at the staging\n"+
+				"cluster's ingress.")
+	}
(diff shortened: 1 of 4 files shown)


CodeStaple:
Right, and it was worse than a wording problem: the help promised something `get` genuinely did not do. Fixed in 53a50f0 by making it true rather than by softening the sentence.

The backend already emits `demo_preview_dns_warning` and the portal already renders it. The CLI never read the field, so the one verdict that explains why a demo URL does not load was being dropped on the floor while `get` printed only the TLS one. It now reads it and prints it **before** the TLS warning: an unresolvable hostname is usually why the certificate complaint is there too, and reading the symptom ahead of the cause sends people to the wrong setting.

It carries a `yaml` tag like its neighbours, so `-o yaml` does not lowercase it into a key that does not exist in the contract — the same trap as the earlier `demobasedomain` bug on this surface.

Three tests: the warning reaches the operator, DNS is ordered before TLS, and the yaml key matches the wire key.

CodeStaple:
Both looked at; one needed a change, one did not.

**[medium] New wire field may break the flag-registration test** — it cannot, and you were right to ask since the body was not in the diff. `TestPreviewSettingFlagsAreRegisteredForEveryWireField` walks `previewSettingFlagFields`, which is the map of **writable set flags** to wire fields, and then the registered `demo-*` flags. It never touches the response struct. `demo_preview_dns_warning` is response-only and is not a flag, so it is outside both loops — the same reason `demo_preview_tls_warning` was never excluded by name. No list to update, and the full suite passes.

**[low] Help promises a DNS verdict older backends never send** — real, and the note is added in 98c5e2b. An absent key decodes to empty, which is indistinguishable from "checked, nothing to report", so against a platform predating the field the silence is *unknown* rather than all-clear.

I left it as version skew rather than modelling it, and wrote down why: the field ships on the hosted platform, and giving it a per-field "could not be determined" state would cost every reader a distinction almost nobody is on the wrong side of. Recording the assumption is the part that was missing — it is exactly the kind of thing that is obvious for a week and invisible for a year.

CodeStaple:
You were right to raise it again — I implemented it in f6de816 rather than defend the tradeoff a second time.

`PreviewDNSReported` records whether the platform sent the field at all, so an absent verdict and an empty one are no longer the same thing. When it was never reported and a preview domain is set, `get` says so instead of leaving silence a reader takes for "your hostnames resolve".

It carries `json:"-" yaml:"-"` — it describes the platform that answered, not the organisation's settings, and the structured output should keep mirroring the wire.

Being straight about why I changed position: my previous reply documented this as an accepted tradeoff, and that was the wrong call *for this lane in particular*. Letting an absent answer pass for a good one is the exact failure the whole PLA-773 thread has been undoing — a credential reporting healthy while 404ing, a truncated body reading as a broken credential, a preview domain reporting nothing while previews fell to plain http. Making an exception to save one bool would have been inconsistent with every other fix in the thread.

Three tests: the caveat appears when nothing was reported, stays away when the platform did report and found nothing, and stays away when there is no preview domain to be unsure about.

CodeStaple:
The first one was a live bug in the commit before it, and you called the mechanism exactly. Both fixed in a96bab3.

**[medium] Absent-vs-empty depends on backend wire behaviour** — confirmed against the platform: `DemoPreviewDNSWarning *string `json:"demo_preview_dns_warning"`` carries no `omitempty`, so a healthy organisation gets an explicit `null` (I checked a live response earlier in this lane and the sibling TLS field comes back as `null` exactly that way). Decoding that into a `*string` yields nil, which is indistinguishable from a key that was never sent — and `PreviewDNSReported` was inferred from precisely that nil. So the caveat would have greeted **every healthy organisation**: a false alarm on the one path that was fine, which is worse than the silence I was trying to fix.

Now decoded as `json.RawMessage`, which is nil only when the key is genuinely absent, then unmarshalled to the string. Three decode-level tests pin `null`, absent, and a real warning apart, since that distinction is the whole point and it lives in the decoder rather than the renderer.

**[low] Help promises specifics the fallback does not deliver** — right, and they are two different messages. Naming the wildcard and the address is the platform's verdict; the fallback is the case where there is no verdict and therefore nothing concrete to name. The help now attributes the specifics to the verdict and describes the fallback as what it is.

Worth saying plainly: this is the second time in two commits that writing down what something claims exposed that it did not do it. Both were mine.

PR #147, opened 2026-08-24
fix(cluster): stop calling deprovision/provision a stop/start (ankra-54il0)

Myothas:
Bead: ankra-54il0
Linear: https://linear.app/ankra/issue/PLA-792

## Why

`ankra cluster deprovision` described itself as **"Deprovision (stop) a managed cluster"**
and its Long help promised it *"will shut down the cluster but not delete it"* — while its
own confirmation prompt three lines later warned that it *"deletes all its cloud resources
(servers, networks, SSH keys)"*. Both cannot be true. The prompt is the accurate one.

Deprovision is not the power lane:

| | lane | what happens to `addon` / `addon_configuration` / `manifest` |
|---|---|---|
| `ankra cluster deprovision` | `DeprovisionProviderCluster` (`cluster` `go/internal/usecase/providers/lifecycle.go:165`) | `MarkAllResourcesStoppingForDeprovision` sets `delete_permanently=true` on **every** resource, then the cloud-termination DAG uninstalls them (`CloudTerminationWorkloadKinds`, `enginekit/clusterengine/cloudtermination.go:37`) |
| `ankra <provider> cluster stop` | `StopProviderCluster` → `RunStopClusterSequence` (`enginekit/clusterengine/lifecyclesequence.go:36`) | `SnapshotClusterStateBeforeStop` writes a pre-stop snapshot; `MarkAnkraStackResourcesDown` only flips state to `down` — rows and definitions preserved in place |

A customer power-cycled a GPU cluster through `deprovision`/`provision` believing it was a
stop/start. The gpu-operator addon values they had patched with `ankra cluster addons
upgrade` (the K3s `CONTAINERD_CONFIG` / `CONTAINERD_SOCKET` paths) came back at chart
defaults and every GPU workload sat `Pending` until they re-ran the upgrade. They asked
explicitly for "a prominent warning at provision time" — this is that warning.

## What changed

Wording only — no behaviour change.

- Both `Short` strings drop the `(stop)` / `(start)` labels.
- `deprovision`'s `Long` says plainly that it is a teardown, that all cloud resources are
  released **and every stack resource is uninstalled**, and points at
  `ankra <provider> cluster stop|start` and `ankra cluster power-schedules` for an actual
  power cycle.
- `provision`'s `Long` says stacks are redeployed from the cluster's stored stack
  definition, and the command now prints that as a note on **stderr** so `--output json`
  stays machine-parseable on stdout.
- The deprovision confirmation prompt says the stack is uninstalled too, and that this is
  a teardown rather than a power-off.

## Tests

New `cmd/reconcile_lifecycle_wording_test.go` — 5 tests, all verified failing on `master`
and passing here (checked by stashing `cmd/reconcile.go` and re-running):

- deprovision help no longer contains `(stop)` / `but not delete it`, and does mention the
  real power lane;
- provision help no longer contains `(start)`;
- the confirmation prompt carries the teardown wording;
- provision emits the "verify in-place patches" note;
- the note is absent from `--output json` (structured output stays clean).

`gofmt` clean on the touched files, `go build ./...`, `go vet ./cmd/`, and
`go test ./cmd/` all pass. `golangci-lint` is not installed on this machine — CI is the
gate for it.

## Not in scope

Whether `provision` *should* restore `addon` / `addon_configuration` / `manifest` the way
it restores infra (`restoreCloudResources` un-deletes only `InfraOnlyKinds` +
`CloudInfraKinds`, and `healStuckResources` only touches rows with `deleted_at IS NULL`)
is a real state-durability question that needs a product decision and prod-side
verification. Tracked separately in **ankra-dic4y**; deliberately not guessed at here.

Linear: https://linear.app/ankra/issue/PLA-792
Bead: ankra-54il0

### DIFF
--- a/cmd/reconcile.go
+++ b/cmd/reconcile.go
@@ -245,8 +245,17 @@ func resolveClusterFromArgs(cmd *cobra.Command, args []string) (string, string, 
 
 var clusterProvisionCmd = &cobra.Command{
 	Use:   "provision [cluster_name]",
-	Short: "Provision (start) a managed cluster",
-	Long: `Start a managed cluster that was previously created but is not yet running.
+	Short: "Provision a managed cluster (build its infrastructure and redeploy its stacks)",
+	Long: `Provision a managed cluster that was created or deprovisioned but is not running.
+
+Provisioning rebuilds the cluster's infrastructure and then redeploys its stack
+resources from the cluster's stored stack definition. Verify anything you patched
+in place afterwards - see "ankra cluster addons list <addon>" and
+"ankra cluster manifests list".
+
+This is not a power-on. To resume a cluster you powered off, use the provider's
+start command (for example "ankra hetzner cluster start") or
+"ankra cluster power-schedules".
 
 If no cluster name is provided, uses the currently selected cluster.`,
 	Args: cobra.MaximumNArgs(1),
@@ -281,6 +290,11 @@ If no cluster name is provided, uses the currently selected cluster.`,
 		if result.MarkedToStartAt != "" {
 			fmt.Printf("  Scheduled at: %s\n", result.MarkedToStartAt)
 		}
+		// Stack resources are redeployed from the stored stack definition, so an
+		// in-place patch is worth re-checking once the cluster is back online.
+		_, _ = fmt.Fprintln(cmd.ErrOrStderr(),
+			"note: stacks are redeployed from the cluster's stored stack definition; "+
+				"verify in-place patches with 'ankra cluster addons list <addon>' once the cluster is online")
 		return nil
 	},
 }
@@ -302,8 +316,19 @@ const (
 
 var clusterDeprovisionCmd = &cobra.Command{
 	Use:   "deprovision [cluster_name]",
-	Short: "Deprovision (stop) a managed cluster",
-	Long: `Stop a running managed cluster. This will shut down the cluster but not delete it.
+	Short: "Deprovision a managed cluster (tear it down; the cluster record is kept)",
+	Long: `Tear down a managed cluster. The cluster record is kept so it can be provisioned
+again later, but everything it runs on is released.
+
+This is a teardown, not a power-off:
+
+  - all cloud resources are released (servers, networks, SSH keys);
+  - every stack resource on the cluster is uninstalled, and a later
+    "ankra cluster provision" redeploys them from the stored stack definition.
+
+To power a cluster off and back on while keeping its state, use the provider's
+stop/start commands (for example "ankra hetzner cluster stop" and
+"ankra hetzner cluster start") or "ankra cluster power-schedules" instead.
 
 If no cluster name is provided, uses the currently selected cluster.
 
@@ -340,7 +365,8 @@ to the provider-specific deprovision endpoint so cloud resources are released.`,
 
 		if err := confirmPrompt(
 			cmd.InOrStdin(), cmd.OutOrStdout(),
-			fmt.Sprintf("Deprovision cluster %q? This deletes all its cloud resources (servers, networks, SSH keys)! [y/N]: ", clusterName),
+			fmt.Sprintf("Deprovision cluster %q? This deletes all its cloud resources (servers, networks, SSH keys) "+
+				"and uninstalls every stack resource on it - this is a teardown, not a power-off. [y/N]: ", clusterName),
 			yes,
 		); err != nil {
 			return err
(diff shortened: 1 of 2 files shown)


Myothas:
**🤖 tech-agent** — ready for review, not merging: this PR's premise was retracted after I opened it

**What this PR does:** wording-only change to `ankra cluster deprovision` / `provision` help in `cmd/reconcile.go`, plus 5 tests pinning it. It drops the `(stop)` / `(start)` labels, describes deprovision as a teardown, points at `ankra <provider> cluster stop|start` and `ankra cluster power-schedules` as the real power lane, and adds a provision-time note on stderr telling operators to re-verify in-place patches.

**Merge-gate status:** 6 of 8 conditions hold — my own PR (marker present), CI green on `27d9e38d` (`lint-test-build` ×2 SUCCESS), no human changes-requested, small (156 lines / 2 files), merges cleanly (`MERGEABLE` / `CLEAN`), originating ticket passed the injection screen. Two fail, so this is ready-for-review rather than merged.

---

### 1. The factual premise no longer holds

I wrote this PR's "Why" section, and the provision-time advisory, on the belief that a deprovision→provision discards in-place `addons upgrade` patches. [cluster#1833](https://github.com/ankraio/cluster/pull/1833) (merged 2026-08-24 13:34Z, `677aa253a`) resolves PLA-792 as **not-a-bug**, on read-only live forensics of the same cluster and the same 11:32Z provision this PR cites:

- `configure_addon` at 11:32Z targeted the same `addon_configuration` row `34ce3111`, with a null delta — no definition change;
- the nvidia-container-toolkit DaemonSet that provision created mounts the **patched** K3s hostPath, which gpu-operator only renders when the patched `CONTAINERD_CONFIG` is in effect;
- `ankra cluster addons values gpu-operator` still carries the patch;
- the `/runtime/config-dir` paths the customer read as chart defaults are the operator's fixed in-container remap mounts.

The addon values **did** survive, and cluster#1833 adds Postgres-backed regression suites pinning that across the deprovision/start sequence. Two things in this PR are therefore wrong or misleading as written:

1. **the new stderr note** — `note: stacks are redeployed from the cluster's stored stack definition; verify in-place patches ...` — fires on *every* successful provision and tells operators to distrust behaviour engineering has just certified as durable and regression-pinned;
2. **deprovision's new `Long`** — "every stack resource on the cluster is uninstalled, and a later `ankra cluster provision` redeploys them from the stored stack definition" — implies the stored definition may differ from the patched state. Per the forensics, the patch *is* the stored declared state.

The part I still believe is correct and worth landing: dropping the `(stop)` / `(start)` labels and pointing at the provider stop/start commands. `deprovision` genuinely releases cloud resources while `stop` keeps them allocated, and the old `Long` ("will shut down the cluster but not delete it") contradicted the command's own confirmation prompt three lines below it. Nothing in cluster#1833 disputes that.

I did not rewrite the wording myself: which of these claims to make to operators depends on forensics I cannot run, and it is precisely the question cluster#1833 just adjudicated.

### 2. The Ankra AI review is incomplete

The review on `27d9e38d` says so itself — *"This review is incomplete. The model's answer was cut off."* I cannot show every finding is resolved when I cannot see them all. Un-drafting did not re-trigger it (head-SHA dedup); a re-review fires automatically on the next push.

Its one visible concern I did check rather than leave dangling:

> an advisory note written to stderr on every successful provision is unconditional in the visible diff and its exclusion from JSON output mode relies on code not shown here

**Not a defect.** The pre-existing guard at `cmd/reconcile.go:257-259` returns before the note is reached:

```go
if format != outputDefault {
    return encodeStructured(cmd.OutOrStdout(), format, result)
}
```

`TestClusterProvisionNoteIsSuppressedForStructuredOutput` pins it. (Moot if the note is dropped per point 1.)

---

### What I would like checked

@Myothas — and Mark Shine, as the author of the PLA-792 verdict:

1. **Should the provision-time advisory and the "stored stack definition" wording be dropped**, given cluster#1833? My read is yes, but it is a customer-facing claim, so I want a human's sign-off rather than my guess.
2. **If so, is the reduced PR still worth landing** — `(stop)`/`(start)` labels removed, teardown-vs-power-lane wording kept, tests trimmed to match? I can push that revision on request.
3. **Outside this PR:** bead `ankra-dic4y` is still OPEN as P1 asserting the revert happened; cluster#1833 supersedes it. (`ankra-gw1ws` has not reached this machine's tracker yet.)

Confidence: high on the merge-gate assessment and the JSON-output answer (both read directly from code and the API); high that the premise changed (read from cluster#1833's merged body). I have not independently verified the prod forensics — no prod access, and I would not use it if I had.

PR #148, opened 2026-08-24
fix(node-group): show availability zones in node-group list (ankra-yg2an)

Myothas:
Bead: ankra-yg2an
Linear: https://linear.app/ankra/issue/PLA-795/1091-az-drift-verified-server-side-general-par-a-and-general-par-b

**Depends on ankraio/cluster#1838** — that PR adds `availability_zones` to the
node-group listing. This one renders it. Merging this first is harmless (the
field is simply absent and the output is unchanged), but it does nothing until
the platform side ships.

## Why

A customer with an OVH 3-AZ cluster found two node groups named `general-par-a`
and `general-par-b` whose nodes all sat in `eu-west-par-c`. Their point about
the CLI, quoted:

> node-group list shows no AZ field at all; only per-node `ovh nodes get` does

That is accurate. Seeing a group's placement currently means one API call per
node, which is why the drift went unnoticed on a cluster explicitly built for
multi-AZ.

## What this changes

- `client.NodeGroupInfo` gains `AvailabilityZones`. Without it the new field is
  dropped by the typed decode even under `-o json`.
- The generic `cluster node-group list` and `cluster ovh node-group list` lines
  print `az=<zones>` for a group that has any.

The zones are **named, not counted**: a count reads `1` both for a group
correctly pinned to one zone and for a group that silently collapsed into the
wrong one — the exact ambiguity this is meant to end.

```
general-par-a         type=b3-32     count=1  labels=0  taints=0  az=eu-west-par-c
ts-par-a              type=r3-128    count=1  labels=1  taints=1  az=eu-west-par-a
```

## Compatibility

The suffix is empty when a group carries no zones, so the list line is
byte-for-byte what it was for every provider without zone placement, for OVH
groups created before zone placement existed, and against a platform that does
not serve the field yet. The three provider-specific listings that can never
carry zones (Hetzner, UpCloud, DigitalOcean) are deliberately untouched.

## Testing

- New unit tests for `nodeGroupZoneSuffix`: empty without zones, the single zone
  named for a collapsed group, all three for a spread one.
- `go test ./cmd/... ./internal/client/...` passes; `go build ./...` clean;
  `gofmt -l` clean on all four touched files.

### DIFF
--- a/cmd/cluster_node_group.go
+++ b/cmd/cluster_node_group.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"os"
 	"strconv"
+	"strings"
 
 	"ankra/internal/client"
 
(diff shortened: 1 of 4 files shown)


Myothas:
@ankra Agreed with your own conclusion on the absent-answer point — recording the verification so the reasoning is on the record rather than only in the review's scratchpad.

**Empty renders nothing, and that is the right direction here.** `AvailabilityZones` empty does conflate three states: the group requested no zone, the provider has no zone placement, and the platform predates the field. Rendering `az=?` for all three would put a question mark on every Hetzner/UpCloud/DigitalOcean line forever, which is noise standing in for information the CLI genuinely does not have. The struct comment names all three cases, so the conflation is documented at the point a reader meets it.

Two things I checked beyond the diff:

**The OVH path really does get the field.** `nodeGroupZoneSuffix` takes `client.NodeGroupInfo`, and `cmd/ovh_cluster.go` passes `ng` from `result.NodeGroups` straight into it — same type, so the compiler is the guarantee here (lint-test-build is green). That matters because the OVH listing is the exact surface the originating ticket (PLA-795) was asking about.

**Coverage of the list surfaces is complete for the behaviour, not just the diff.** There are five node-group list lines in the tree — `cluster_node_group.go`, `ovh_cluster.go`, `hetzner_cluster.go`, `upcloud_cluster.go`, `digitalocean_cluster.go` — and this PR patches two. That is deliberate rather than a gap: the platform only populates `availability_zones` when the engine has an availability-zone field, which today is OVH alone (`OvhTopology()` is the sole assignment of `AvailabilityZoneField` in cluster). The three unpatched commands are provider-specific entry points for providers with no zone placement, so `nodeGroupZoneSuffix` would return `""` on every line — adding the call there would be dead code that only looks like coverage.

The two patched surfaces are precisely the ones that can reach an OVH cluster: the provider-agnostic `cluster node-group list` and the OVH-specific `ovh node-group list`.

The platform half (ankraio/cluster#1838) is merged and deployed, so the field is being served. The `omitempty` on both sides means the ordering was safe either way — against an older platform the suffix is simply absent.

PR #165, opened 2026-08-25
docs(skills): ankra-cicd routes application repositories to the generator first (ankra-ryfuy.8)

Myothas:
## What

The ankra-cicd skill's description now names `ankra application add` as the first stop for an application repository, and a pointer directly under the intro sends the reader to the existing "When Ankra generates the pipeline for you" section and the `ankra-applications` skill before the manual pipeline shapes. reference.md carries the same pointer above its copy-paste examples.

Previously the description routed agents straight into hand-rolling what the platform generates, and the generates-for-you section sat below every manual example - an agent reading top-down had built the pipeline before learning it never needed to.

Complements the skills overhaul that shipped with v0.13.0 (the ankra-applications skill itself); this is only the routing residue.

## Validation

`go build ./...` and `go test ./internal/skills/` green.

Bead: ankra-ryfuy.8


Myothas:
Verified live, not inferred: the ankra-notes onboarding (2026-08-24/25 E2E) produced exactly this set - the generated build-and-publish workflow carries the Semgrep/Checkov/Trivy gates with their scan-artifact contract, publishes immutable sha-<commit> tags, logs in with the Ankra-managed registry secrets, and the setup PR includes the manifests and .ankra/ankra.yaml; pushes to the tracked branch then built and rolled out with no further setup. The wording also matches the ankra-applications skill this points at.

PR #170, opened 2026-08-26
fix(kubeconfig): resolve a cluster across your organisations instead of a misleading 404 (ankra-btmyn)

Myothas:
## What & why

Bead: ankra-btmyn

`kubectl` against an Ankra context failed with a misleading 404 whenever the
CLI's selected organisation was not the one that owns the cluster:

```
Error: kube token request failed: status 404, body: {"detail":"Cluster not found"}
... executable ankra failed with exit code 3
```

The cluster exists, the caller holds a grant on it, and the kubeconfig even
embeds the cluster UUID in the server URL. The platform resolves a cluster
reference *inside the organisation the request is scoped to*, and the exec
credential plugin passes no organisation at all, so the id resolved inside the
selected organisation and came back as a bare 404 that blames the *cluster*.
On a machine that works across organisations that sends you to check access
grants and RBAC for what is an organisation-selection problem. `ankra cluster
access grant` had the same behaviour and needed an explicit `--org`.

This PR does both halves the bead asks for.

**Resolution.** A cluster id the scoped organisation does not have is now
looked up across the organisations the caller belongs to, and the request runs
against the one that owns it (`cmd/cluster_org_scope.go`). Only ids are
searched: cluster names are unique only within an organisation, so resolving
one across them would mean guessing which same-named cluster was meant. The
lookup re-scopes **this invocation only** and never touches the persistently
selected organisation, so parallel sessions in other workspaces are
undisturbed.

**Self-contained kubeconfig entries.** `ankra cluster kubeconfig add <id>`
does the same lookup, so an id from another organisation writes a context
pinned to the *real* owner. Previously that path fell back to the selected
organisation, i.e. it baked the wrong `--org` into the exec args and produced
a context that 404s on first use. Entries that already pin `--org` are
unaffected, and entries written before pinning existed now work through the
resolution above.

**Diagnostics.** When the cluster really is nowhere, the error names the
organisation the request was scoped to and the ones that were searched
instead of repeating `Cluster not found`. A 404 from the mint itself is
decorated the same way, and when the owner is found it says so:

```
kube token request failed: status 404, body: {"detail":"Cluster not found"}
cluster 1111...1111 is not in organisation "Depict" (org-depict); it belongs to "Ankra AB" (org-ankra).
Retry scoped to the owning organisation:
  ankra --org org-ankra <command>   (or ANKRA_ORG=org-ankra)
Or re-add the context so it pins the organisation itself and survives 'ankra org switch':
  ankra cluster kubeconfig add 1111...1111
```

`ankra cluster access` list/grant/revoke get the same resolution and print a
one-line note (stderr, so `-o json|yaml` stays parseable) naming the
organisation they ran against. `kube-token` stays silent: kubectl re-runs the
credential plugin on every command, so a note there would be per-invocation
noise.

**Not regressed:** `ANKRA_ORG=<org> kubectl ...` short-circuits the lookup
entirely (root resolves it into the client override before the command runs,
the first by-id lookup then succeeds, and no organisation list is fetched).
Nothing calls `ankra org switch`.

**Failure-mode care:** a search that could not run proves nothing about a
cluster, so an unreachable organisation list leaves the id to the backend
exactly as before, and a single bad lookup is re-checked rather than promoted
into "this cluster does not exist".

## Test plan

`cmd/cluster_org_scope_test.go` models the backend behaviour that makes the
bug possible (every reference resolves inside the scoped organisation, 404
otherwise) and pins:

- a cross-organisation id resolves and the mint then runs in the owning organisation;
- the credential-plugin path writes nothing to stderr, the `cluster access` path does;
- an in-scope cluster costs no extra calls and leaves the scope untouched;
- an explicit `--org`/`ANKRA_ORG` override survives and skips the search (the documented workaround);
- an id no organisation has names the scoped organisation and every organisation searched, and exits 3;
- an unreachable organisation list passes the id through instead of failing;
- one bad lookup does not become "not found";
- a 404 from the mint names the owning organisation and keeps the original error in the chain (exit code unchanged);
- 403 still gets the access-grant suggestion, transport errors still pass through untouched;
- `kubeconfig add` with a cross-organisation id writes `--org <owner>` and the real cluster name.

The resolution and kubeconfig tests were each confirmed to fail with their
half of the change reverted.

Gates run locally: `gofmt -l` clean on touched files, `go test -race -count=1
-timeout=120s ./...` green, `golangci-lint run` 0 issues (also via the
pre-commit hook).

## Docs checklist

- [x] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); `ankra cluster kube-token`'s `Long` help now describes the cross-organisation lookup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,26 @@
   fixing keys, and `delete` confirms first (or takes `--yes`). `get` and
   `list` take a vault name as well as an id and support `-o json|yaml`.
 
+### Fixed
+
+- **`kubectl` against an Ankra context no longer fails with "Cluster not
+  found" when your selected organisation is not the one that owns the
+  cluster.** The platform resolves a cluster reference inside the
+  organisation the request is scoped to, so a cluster you have a grant on,
+  whose id is right there in the kubeconfig server URL, came back as a bare
+  404 that blamed the cluster - sending you to check access grants and RBAC
+  for what was an organisation-selection problem. A cluster id the selected
+  organisation does not have is now looked up across the organisations you
+  belong to and the request runs against the one that owns it, so contexts
+  written before `--org` was pinned into the exec args keep working across
+  an `ankra org switch`. `ankra cluster kubeconfig add <id>` and
+  `ankra cluster access` do the same lookup, so an id from another
+  organisation writes a context pinned to the real owner instead of a
+  context that 404s on first use. Your selected organisation is never
+  changed, and `ANKRA_ORG=<org> kubectl ...` still short-circuits the
+  lookup. When the cluster genuinely is nowhere, the error now names the
+  organisation the request was scoped to and the ones that were searched.
+
 ## v0.13.0 — 2026-08-25
 
 Promotes v0.13.0-rc0 through rc4 — the kubectl-shaped read surface
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,7 +31,10 @@
   an `ankra org switch`. `ankra cluster kubeconfig add <id>` and
   `ankra cluster access` do the same lookup, so an id from another
   organisation writes a context pinned to the real owner instead of a
-  context that 404s on first use. Your selected organisation is never
+  context that 404s on first use. The cluster from `ankra cluster select`
+  gets the same treatment, since that selection outlives the organisation
+  chosen alongside it: omitting `--cluster` no longer fails where passing
+  that very same id succeeds. Your selected organisation is never
   changed, and neither is an organisation you pinned yourself: `--org` and
   `ANKRA_ORG` are honoured as given, so a cluster that is not in the one you
   named is reported rather than quietly fetched from somewhere else. When
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,9 +32,13 @@
   `ankra cluster access` do the same lookup, so an id from another
   organisation writes a context pinned to the real owner instead of a
   context that 404s on first use. Your selected organisation is never
-  changed, and `ANKRA_ORG=<org> kubectl ...` still short-circuits the
-  lookup. When the cluster genuinely is nowhere, the error now names the
-  organisation the request was scoped to and the ones that were searched.
+  changed, and neither is an organisation you pinned yourself: `--org` and
+  `ANKRA_ORG` are honoured as given, so a cluster that is not in the one you
+  named is reported rather than quietly fetched from somewhere else. When
+  the cluster genuinely is nowhere, the error names the organisation the
+  request was scoped to and the ones that were searched - and only the ones
+  that actually answered, so a lookup that failed is never reported as an
+  absence.
 
 ## v0.13.0 — 2026-08-25
 
--- a/cmd/cluster_access.go
+++ b/cmd/cluster_access.go
@@ -44,7 +44,7 @@ var clusterAccessListCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List access grants for a cluster",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		clusterID, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
+		clusterID, _, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
 		if err != nil {
 			return err
 		}
@@ -83,7 +83,7 @@ admin, cluster-admin.`,
 		if err := validateAccessRole(accessRoleFlag); err != nil {
 			return err
 		}
-		clusterID, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
+		clusterID, _, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
 		if err != nil {
 			return err
 		}
@@ -128,7 +128,7 @@ or an email address to revoke every grant that member has on the cluster.`,
 	Args: cobra.ExactArgs(1),
 	RunE: func(cmd *cobra.Command, args []string) error {
 		target := args[0]
-		clusterID, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
+		clusterID, _, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
 		if err != nil {
 			return err
 		}
--- a/cmd/cluster_access.go
+++ b/cmd/cluster_access.go
@@ -44,7 +44,7 @@ var clusterAccessListCmd = &cobra.Command{
 	Use:   "list",
 	Short: "List access grants for a cluster",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		clusterID, err := resolveKubeTokenClusterID(accessClusterFlag)
+		clusterID, err := resolveGatewayClusterID(accessClusterFlag, os.Stderr)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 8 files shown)


Myothas:
@ankraai All three findings addressed in 71163fc — the medium one was a real instance of the exact confusion this PR exists to remove, thanks.

**[medium] Failed per-organisation lookups recorded as "searched, not found".** Fixed properly rather than by blanket-marking any error inconclusive: `GetClusterByID` previously returned a plain error for both "the response came back and had no match" and "the lookup failed", so the loop had nothing to discriminate on. `internal/client` now exports an `ErrClusterNotFound` sentinel (matching the existing `ErrAddonNotFound` / `ErrVariableNotFound` house style) that both cluster lookups wrap. The search loop appends to `searchedLabels` only on a clean miss; any other error is kept as `firstFailure` and, if nothing matched, becomes `search.err` — which every caller already treats as "leave the id to the backend". Pinned by `TestFindClusterInOtherOrganisationsNeverCountsAFailedLookupAsAMiss`: all lookups fail, so `searchedLabels` is empty, the search is inconclusive, and the command still succeeds.

**[low] `kubeconfig add` writing a guaranteed-broken raw context.** Fixed: the caller now checks `search.err` as well as the adopted flag. A search that ran cleanly and settled the question errors out with the organisations named (exit 3, no file written); the passthrough survives only for the inconclusive case, which is the pre-existing behaviour. Pinned by `TestKubeconfigAddRefusesToWriteAContextItKnowsIsBroken`, and `TestKubeconfigAddTreatsUUIDFlagAsID` still covers the inconclusive passthrough.

**[low] The `ANKRA_ORG` claim overstating what happens.** You were right that the code was the weaker half of this — a stale or mistyped explicit override was silently replaced. Took the first option: `resolveOwningOrganisation` gates adoption on there being no explicit `--org` / `ANKRA_ORG`. Behind a pinned organisation the search still runs, but only to name the owner and the exact retry in the error, so an explicit pin is honoured as given. The CHANGELOG now says that instead of claiming a short-circuit. Pinned by `TestResolveGatewayClusterIDNeverReplacesAnExplicitOrganisation` (override pinned to an organisation that does *not* own the cluster), alongside the existing "override happens to be correct" case.

Myothas:
@ankraai Both addressed in b51a4e4. You were right on (1), and right that it was the *same* finding again — so this round fixes the representation rather than the third call site.

**[medium] A failed retry reported as "cluster does not exist".** Confirmed, and while generalising I found a fourth instance you hadn't flagged yet: `resolveKubeconfigTargets` had its own copy of the in-scope lookup whose failure fell straight into the definitive not-found branch too.

The root cause is representational, not local. `GetClusterByID` returns `(cluster, error)`, which spells "the backend said no" and "the backend said nothing" identically, so every call site had to classify the error itself — and `err != nil` reads naturally as absence at all of them. Three sites classified, two got it wrong.

So there is now one tri-state primitive:

```go
func (l clusterLookup) found() bool  { return l.present }
func (l clusterLookup) absent() bool { return !l.present && l.unanswered == nil }
```

`found()` and `absent()` are deliberately not each other's negation — a lookup that never completed is neither. `lookUpClusterInScope` is the only place permitted to turn a `(cluster, error)` pair into a judgement about existence, so `errors.Is(..., ErrClusterNotFound)` now appears exactly once in the whole flow instead of once per site. One level up, `resolveClusterOrganisation` carries the same discipline for the organisation question (`settled()` / `absent` / neither) and is the single door both `kube-token` and `kubeconfig add` go through — which is what collapsed the fourth instance without a separate fix.

Pinned by `TestLookUpClusterInScopeSeparatesAbsenceFromSilence` (the primitive's three states directly), `TestResolveGatewayClusterIDNeverCallsAFailedRetryAnAbsence` (your exact case: both in-scope lookups fail, the search answers cleanly, and the id still reaches the backend), and `TestKubeconfigAddDoesNotCallASilentOrganisationAnAbsence` (the fourth site). I checked all three fail if the retry branch is collapsed back to two states.

**[low] Mint-404 always sweeping.** Fixed. Half of it was already there — `findClusterInOtherOrganisations` returns before `ListOrganisations` for a non-UUID reference, so the `isLikelyClusterID` gate you suggested exists — but you're right that for a UUID it swept unconditionally, including when resolution had *already* settled the organisation, where the sweep cannot change the answer. `resolveGatewayClusterID` now returns whether the owning organisation ended up known, and `decorateKubeTokenError` skips the sweep when it is. Settled means: the name resolved in scope, the id was confirmed in scope, or the owner was adopted. Unsettled (search inconclusive, or a cluster from the local selection file, which can name another organisation's cluster) still sweeps, because that is the case where the message is the whole point.

Pinned by `TestDecorateKubeTokenErrorSkipsTheSweepWhenTheOrganisationIsKnown` (asserts zero org listings and zero lookups when known, and that the owner is still named when not) and `TestResolveGatewayClusterIDReportsWhetherTheOrganisationIsSettled` across all four resolution paths.

Myothas:
@ankraai Fixed in 78147ea rather than defended — the fallback wasn't deliberate, it was an oversight, and you framed it exactly right: same identifier, same user, different outcome depending only on how it was supplied.

The reasoning that settles it: the selection is a local file written by `ankra cluster select`, so it long outlives the organisation that was selected alongside it. That makes it *exactly* as likely to name another organisation's cluster as an id typed at the prompt — arguably more so, since it can sit there for weeks — and there is no reason to trust it more than the flag. It now goes through `resolveClusterOrganisation` on the same terms, so the no-flag and explicit-id paths agree. This reaches `ankra cluster access` list/grant/revoke too, which default to the selected cluster.

Worth noting what this does *not* cost: the kubectl credential-plugin path always passes `--cluster` (that is what `kubeconfig add` writes), so the extra lookup lands only on interactive invocations, where the explicit-id path already paid it.

`TestResolveGatewayClusterIDTreatsTheSelectionLikeATypedID` pins it as a **comparison** rather than an assertion about the no-flag path in isolation — the bug was the disagreement, so the test resolves the same cluster both ways and asserts the id, the settled flag, and the resulting organisation scope all match. Reverted, it fails with precisely the asymmetry you described:

```
no-flag path = ("1111...", false), explicit path = ("1111...", true); the same id must resolve the same way
no-flag organisation scope = "", explicit = "org-ankra"
token mint after adoption failed: status 404, body: {"detail":"Cluster not found"}
```

That is the last open finding from my side. CHANGELOG updated to match.

Myothas:
@ankraai Thanks — and your first observation turned out to understate it, so I acted on it in df4f61d.

`resolveKubeTokenClusterID` wasn't just a wrapper discarding the `settled` flag: once `kube-token` and `cluster access` both moved to `resolveGatewayClusterID`, **nothing in the binary called it at all**. It survived only because five tests still did, which is exactly the shape that keeps dead code alive and green. Deleted, with those tests moved onto `resolveGatewayClusterID(ref, nil)` — which is what the quiet path actually is. The two tests named after the wrapper are renamed for the behaviour they cover rather than the function they call, so they don't go stale the next time the entry point moves. No behaviour change; the credential plugin's silence is still pinned by `TestKubeTokenPathStaysQuiet`.

Your second observation is right and deliberate: `cluster access` passes `os.Stderr` so interactive users see the re-scoping note, while `kube-token` passes nil because kubectl reruns it on every command.

One thing worth flagging for whoever reads this thread: **the review comment currently rendered on the PR is truncated** — it carries the `Reviewed commit 78147ea0bf6f` footer and the "This review is incomplete" banner, with the body degenerating into repeated `Q` tokens. That looks like a generation failure on your side rather than anything about the diff; the substantive sign-off is in your reply comment above it, not in the review block. This push should produce a clean one on `df4f61d`.

PR #211, opened 2026-08-31
feat(mesh): ankra cluster mesh make-ready (ankra-0xsdd.25.6)

CodeStaple:
Bead: ankra-0xsdd.25.6

`ankra cluster mesh make-ready <cluster_id> [--site-public-ip <address>]` — the CLI half of cluster#2269. Turns an existing cluster mesh-capable in place: the platform allocates its Cilium identity and overlay range day-2, stamps the overlay onto its stored definitions, and sets its resources converging so every node joins the WireGuard overlay. Prints the identity and how many resources are converging (or renders structured with `-o json`); refusals surface the platform's reason verbatim (unwired provider, non-kubeadm, missing site address).

Tests: the command passes cluster and site address through, and a platform refusal fails the command carrying its reason. Full `cmd` + `client` suites and the repo linter green via pre-commit.

### DIFF
--- a/cmd/cluster_mesh.go
+++ b/cmd/cluster_mesh.go
@@ -135,6 +135,36 @@ var clusterMeshLeaveCmd = &cobra.Command{
 	},
 }
 
+var clusterMeshMakeReadySiteIP string
+
+var clusterMeshMakeReadyCmd = &cobra.Command{
+	Use:   "make-ready <cluster_id>",
+	Short: "Turn an existing cluster mesh-capable without recreating it",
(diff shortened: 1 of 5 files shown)


CodeStaple:
Both findings handled:

1. **Decode-miss as success** — fixed in ae057b6: a response whose `make_ready` object is absent (zero id *and* empty name — an allocated identity always carries both) errors with "the platform answered without a make_ready result" instead of printing zero values.

2. **Package-level flag variable** — this is the file's and repo's cobra convention (every `Flags().StringVar(&…)` in `cmd/` binds a package-level var); a one-off closure-scoped flag here would be the inconsistency. CLI commands run once per process, so cross-invocation sharing has no live path.

PR #215, opened 2026-08-31
fix(clusters): cluster info finds clusters past the first page (ankra-99r3d)

CodeStaple:
`ankra cluster info <name>` sent `cluster_name=`, a query parameter the server never read (`listClustersHandler` reads `cluster_id/name/state/kind/...`). The request silently returned the default first page of 25 and the CLI filtered client-side - so for an organisation with more than 25 clusters, any cluster sorted past page 1 answered `no cluster found for name`.

Fix: narrow with the server's real `name` filter + explicit `page=1&page_size=100`, exact match client-side - the same truncation fix `listAllClusters` already carries. Fake-server test updated to pin the real query shape (rejects `cluster_name=`, requires the explicit page size).

Found during the playground/managed-k8s review (epic ankra-rnzyx).

Bead: ankra-99r3d

cc @ankraai

https://claude.ai/code/session_01Sj8otCXmLWGAjdaHJFVj7T

### DIFF
--- a/internal/client/clusters.go
+++ b/internal/client/clusters.go
@@ -78,11 +78,16 @@ func (c *Client) ListClusters(page int, pageSize int) (*ClusterListResponse, err
 	return response, nil
 }
 
-// GetCluster looks up a cluster by exact name. The backend's
-// /api/v1/clusters?cluster_name=... mode returns a ClusterListResponse
-// with the matching row in `result` (or an empty result on no match).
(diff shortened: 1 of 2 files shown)


CodeStaple:
@ankraai Taken - GetCluster now walks every filtered page (same bounded loop shape as listAllClusters) and only answers not-found once the pages are exhausted; a new fake-server case pins the exact match landing on page 2 of the substring-filtered results.

PR #216, opened 2026-09-01
feat(application): ankra application ship — one-shot idea-to-live (ankra-lhsv6, + login status, git credentials in list, flag consistency)

CodeStaple:
Beads: ankra-lhsv6 ankra-k9my4 ankra-zuvv0 ankra-sjf1u

Four dogfood findings from the idea-to-live flow, shipped together because the headline command is what surfaced the other three.

## ankra-lhsv6 — `ankra application ship [path]`

One command from a local checkout to a live deployment, composing the lanes that already exist as separate subcommands (`add`, `get`, `workflow-runs`, `build`, `deploy`, `installations`, pods/ingress reads):

1. **Resolve or register** — matches the checkout's repository against the applications listing, with `--name` as part of the application's identity: a repository hosting several applications ships the named one, `--name` with a new name registers another rooted at the given path, and a multi-application repository without `--name` is refused with the names listed (the fourth live finding: matching by repository alone adopted the OTHER application and redeployed it onto the caller's hostname). Otherwise registers via the extracted `application add` flow (ship carries the identical add flag set, enforced by a test).
2. **Wait for setup** — polls the application to state `up`; failure states surface `error_message` and exit non-zero. The setup PR URL is reported as soon as it exists.
3. **Merge gate** — an unmerged setup PR is printed clearly and polled at human speed (15s) until merged. With `--ankra-build` the gate and repo CI are skipped: the tracked branch's head commit (from the branches read) is built on Ankra's builders, reusing the build lane's own polling; the feature-flag 404 is translated to "platform builds are not enabled for this organisation".
4. **Wait for an image** — polls workflow-runs for the tracked branch until the latest run concludes; a failed conclusion prints the run URL and exits non-zero.
5. **Deploy** — `--cluster` or the selected cluster; the resolved organisation NAME and cluster NAME are echoed before acting (org-drift defence), and a selected cluster from another organisation is refused. `--namespace` defaults to the platform's own name derivation (mirrored client-side so the verify/URL reads target the right namespace). Only user-supplied `--set` inputs are passed through; image tag and ingress stay platform defaults. Supplied `--set` values ALWAYS produce a deploy: an in-flight deploy is waited to settlement first and even a healthy installation is deployed again to apply them — adoption without a new deploy is reserved for flag-less resumes (the first live run adopted a replicas=0 park deploy and silently dropped the caller's values).
6. **Verify** — follows the installation to `healthy`; a settled `failed` surfaces the platform execution error plus the namespace's Warning events and exits non-zero. The Live line is then gated on pods actually Running and Ready: installation health is vacuously satisfied at zero replicas, so a namespace verified at zero running pods reports the deployment as **parked** (state `parked` in `-o json`) and exits non-zero instead of presenting a 503 URL as live.
7. **URL** — reads the deployments surface first: each deployment row carries the platform-derived `ingress_host` with its `ingress_host_publication` verdict, and a `publisher_absent` verdict is surfaced as a caveat instead of presenting an unresolvable URL as live. Rows without a hostname fall back to reading the namespace's Ingress resources. Prints `Live: https://<host>` as the final stdout line, or says the app deployed unpublished (with the deploy result's derivation message when the platform provided one).

`-o json` emits one structured result document (application id, setup PR, build ref, installation id, namespace, url, state); all progress goes to stderr so stdout stays parseable. **Idempotence is the core design**: every step reads where the flow actually is before acting — a re-run reuses the registered application, skips a healthy installation, rejoins an in-flight deploy, and an expired `--timeout` exits 5 with exactly that resume hint.

**URL sourcing (corrected during review prep):** the platform's deployments read (`GET /api/v1/org/applications/<id>/deployments`) does expose the published hostname — each row carries `ingress_host` paired with the four-valued `ingress_host_publication` verdict (`not_applicable` / `publisher_present` / `publisher_absent` / `unknown`). An earlier revision of this PR claimed otherwise from a stale checkout of the backend. Ship now resolves the URL from that row first (the CLI passes deployments through as raw JSON, so no client model needed extending), keeps the namespace Ingress read as the fallback for rows without a hostname (hand-installed add-ons, older platforms), and keeps the honest "no published URL" path.

### Usage transcript (real command code against the scripted platform, from the test harness)

```
$ ankra application ship . --cluster production
Shipping to organisation "Acme Org", cluster "production".
Registered application "shop" (b7f4c1de-2a03-4d61-9f8e-51c0a6d3e742) for acme/shop@main with credential "github-acme".
Waiting for application setup (state: creating) - follow it with 'ankra application get b7f4c1de-2a03-4d61-9f8e-51c0a6d3e742'.
Setup pull request: https://github.com/acme/shop/pull/1
Application setup is complete (state: up).
Waiting for the setup pull request to be merged:
  https://github.com/acme/shop/pull/1
Merge it to let the repository's CI build the first image (or re-run with --ankra-build to build on Ankra's builders instead).
Setup pull request merged.
CI is building the image: https://github.com/acme/shop/actions/runs/7
CI built the image: https://github.com/acme/shop/actions/runs/7
Deploy started (installation 1a2b3c4d-3333-4444-8555-666677778888, namespace "shop").
Deploy started. Track progress under Deployments.
Waiting for the workload to become healthy (installation status: deploying) - watch pods with 'ankra cluster get pods -n shop'.
Installation is healthy.
Pods in namespace "shop": 2 running and ready of 2.
Live: https://shop.acme.ankra.cc
```

## ankra-k9my4 — `ankra login status`

`ankra login status` used to fall through to plain `ankra login` and silently start a browser auth flow. Now: unknown positional arguments under `login` are rejected (exit 2) with a pointer at `ankra login status`, and the new subcommand reports logged in or not, the API base URL in use, the token source (config file / env / flag, plus the saved token name), and the current organisation (reusing the org-current read) — never touching a browser. Exit 6 when not logged in or the token is rejected, `-o json` supported. Verified against the live platform:

```
$ ankra login status
Logged in:    yes
Base URL:     https://platform.ankra.app
Token source: config (/Users/mash/.ankra.yaml)
Token name:   ankra-cli-…-20260716-114206
Organisation: Ankra - Development (2128a484-…)
```

## ankra-zuvv0 — git credentials in `credentials list`

The portal and the CLI both list from `GET /api/v1/org/credentials`; the current backend includes GitHub credentials in the unfiltered listing, but the reported org saw them omitted. The fix makes the CLI robust either way: when the unfiltered listing contains no github rows, they are fetched through the exact provider-filtered read `application add` already resolves its credential from (`ListCredentials(provider=github)`) and merged by id — a backend that already includes them is untouched (no second read, no duplicates). App-installation-backed credentials render as `github (App)`, and the table points at `ankra credentials repositories <name>` (the table has no description column, so the cross-reference is a line under it). The bead's "no `credentials github` subcommand" note is not addressed here — `credentials list --provider github` + `credentials repositories` cover discovery.

## ankra-sjf1u — flag consistency

- `cluster get resources -A`: the shorthand already exists on master (registered since v0.2.0 and verified on the built binary); a test now locks the `-A` contract in across `get pods`, `get resources`, and every namespaced kind command so it cannot regress.
- `ankra cluster domain` with no positional argument now uses the selected cluster (or the inherited `--cluster` flag) like its sibling commands, instead of erroring "accepts 1 arg(s)"; the resolved cluster is named on stderr because `--enable`/`--remove` mutate through the same fallback.
- Skipped from the bead (out of this PR's scope): a `get ingress` kind alias (`get ingresses` exists) and the `cluster info` stale-selection reselect prompt — though ship itself now refuses a selected cluster from another organisation.

## Tests & gates

`go build ./...`, `go test -race -count=1 ./...` (all packages green), `golangci-lint run` (0 issues), pre-commit gate passed. New coverage: ship state machine (register+deploy+URL, idempotent re-run, setup failure, failed workflow, `--ankra-build` merge-gate skip + feature-flag 404, failed deploy surfacing error+events, resumable timeout exit 5, cross-org refusal, structured output, namespace derivation, URL from the deployments row, publisher_absent caveat, namespace-Ingress fallback), login status (unknown-arg guard, not-logged-in exit 6, env/config token sources, rejected token, JSON), credentials merge (merge, no-double-read, filter passthrough, JSON), domain fallback (selected cluster, no-selection error), `-A` shorthand consistency.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,8 +13,13 @@
   to the target cluster, and follows the installation until the workload is
   healthy - then prints `Live: https://<host>` from the deployment's
   published ingress hostname (with a caveat when nothing on the cluster
-  publishes DNS for it). Re-running ship is safe at any point: it re-reads
-  where the flow
+  publishes DNS for it). The Live line is gated on pods actually running and
+  ready: a workload scaled to zero reports "parked" and exits non-zero
+  instead of presenting a 503 URL as live. Supplied `--set` values always
+  produce a deploy - an in-flight deploy is waited to settlement and even a
+  healthy installation is deployed again to apply them; adoption without a
+  new deploy is only for flag-less resumes. Re-running ship is safe at any
+  point: it re-reads where the flow
   actually is and continues from there, and an expired `--timeout` exits 5
   with exactly that hint. `--ankra-build` builds the first image on Ankra's
   builders instead of waiting for the merge and the repository's CI, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,8 +11,10 @@
   waits for the setup analysis, walks you through the setup pull request
   merge gate, waits for a green image build of the tracked branch, deploys
   to the target cluster, and follows the installation until the workload is
-  healthy - then prints `Live: https://<host>` when an ingress publishes
-  one. Re-running ship is safe at any point: it re-reads where the flow
+  healthy - then prints `Live: https://<host>` from the deployment's
+  published ingress hostname (with a caveat when nothing on the cluster
+  publishes DNS for it). Re-running ship is safe at any point: it re-reads
+  where the flow
   actually is and continues from there, and an expired `--timeout` exits 5
   with exactly that hint. `--ankra-build` builds the first image on Ankra's
   builders instead of waiting for the merge and the repository's CI, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,47 @@
 # Ankra CLI Changelog
 
+## Unreleased
+
+### Added
+
+- **`ankra application ship [path]`** takes a local checkout to a live
+  deployment in one command, composing the lanes that already exist as
+  separate subcommands: it registers the repository as an application when
+  the organisation does not have it yet (same flags as `application add`),
+  waits for the setup analysis, walks you through the setup pull request
+  merge gate, waits for a green image build of the tracked branch, deploys
+  to the target cluster, and follows the installation until the workload is
+  healthy - then prints `Live: https://<host>` when an ingress publishes
+  one. Re-running ship is safe at any point: it re-reads where the flow
+  actually is and continues from there, and an expired `--timeout` exits 5
+  with exactly that hint. `--ankra-build` builds the first image on Ankra's
+  builders instead of waiting for the merge and the repository's CI, and
+  `-o json` emits one structured result document for scripts.
+- **`ankra login status`** reports whether you are logged in - the API base
+  URL in use, where the token came from (config file, environment, or flag),
+  the saved token name, and the current organisation - without ever opening
+  a browser, and exits 6 when not logged in so scripts can probe the state.
+
+### Fixed
+
+- **`ankra login <anything>` no longer starts a browser sign-in.** An
+  unknown positional argument used to fall through to the interactive login
+  flow, so a probe like `ankra login status` on an older CLI silently opened
+  the browser; unknown arguments are now rejected with a pointer at
+  `ankra login status`.
+- **`ankra credentials list` always shows the organisation's GitHub
+  credentials.** When the unfiltered listing omits git credentials, they are
+  now fetched through the same provider-filtered read `application add`
+  resolves its credential from and merged in, App-installation-backed
+  entries read as `github (App)`, and the listing points at
+  `ankra credentials repositories <name>` for what a credential can reach -
+  the GitHub connection is the first prerequisite of `application add` and
+  was invisible.
+- **`ankra cluster domain` without an argument uses the selected cluster**,
+  like its sibling cluster commands, instead of failing with
+  "accepts 1 arg(s)"; the resolved cluster is named on stderr since
+  `--enable`/`--remove` mutate through the same fallback.
+
 ## v0.14.0 — 2026-08-31
 
 Promotes v0.14.0-rc0 through rc8. The headline is the migration lane:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,11 @@
   deployment in one command, composing the lanes that already exist as
   separate subcommands: it registers the repository as an application when
   the organisation does not have it yet (same flags as `application add`),
+  with --name as part of the application's identity - a repository hosting
+  several applications ships the named one, `--name` with a new name
+  registers another rooted at the given path, and a multi-application
+  repository without --name is refused with the names listed rather than
+  one picked silently -
   waits for the setup analysis, walks you through the setup pull request
   merge gate, waits for a green image build of the tracked branch, deploys
   to the target cluster, and follows the installation until the workload is
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -41,6 +41,7 @@ func newApplicationCommand() *cobra.Command {
 		Long:    "Connect application source repositories to Ankra for analysis, packaging, and deployment.",
 	}
 	applicationCommand.AddCommand(newApplicationAddCommand())
+	applicationCommand.AddCommand(newApplicationShipCommand())
 	registerApplicationResourceCommands(applicationCommand)
 	return applicationCommand
 }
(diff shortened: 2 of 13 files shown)


CodeStaple:
Addressed the review — two fixes in 91ec8ef, two answered here.

- **[medium] Global apiClient not rebuilt with resolved credentials** — not changed, deliberately. The `if apiClient == nil { apiClient = newAPIClient() }` guard is byte-for-byte the wiring `persistentPreRunE` does for every auth-requiring command (cmd/root.go), and it is the only production path that ever sets `apiClient` — `login status` is annotated auth-free precisely so that pre-run doesn't run, which means in a real invocation the client is always nil here and always built from the freshly resolved token/baseURL. The "already initialized with different credentials" case only exists when tests inject a mock, and an unconditional rebuild would break exactly that injection (the whole cmd suite's mocking convention). If a second production wiring path ever appears, this and persistentPreRunE would both need the same rework.
- **[medium] Ship drops add-time flags once an existing application is matched** — fixed in 91ec8ef: explicitly-passed registration-shaping flags (`--name`, `--credential`, `--registry-*`) against an already-registered application are now reported as ignored, with pointers at `ankra application credential set` / `ankra application registry set`, and a separate note when the application's tracked branch differs from the checkout's branch. Covered by `TestApplicationShipWarnsAboutIgnoredRegistrationFlags`.
- **[low] Installation status matching is a closed, hardcoded set** — fixed in 91ec8ef: `ensureApplicationDeployed` now deploys only over a missing installation or a settled `failed`/`degraded` one; `healthy` is left alone, and everything else — `pending`, `deploying`, and any status this CLI does not know — is followed as in-flight, so a platform that grows a new transitional state cannot turn a re-run into a racing second deploy. Covered by `TestApplicationShipFollowsAnUnknownInstallationStatus`.
- **[low] Selected-cluster fallback error contract** — confirmed: `resolveActiveCluster`'s no-selection error is `errNoClusterSelected` (cmd/reconcile.go), whose message is "no cluster specified and none selected; pass --cluster <name|id>, or run `ankra cluster select <name>` first" — it names `ankra cluster select` exactly as `TestRunClusterDomain_WithoutAnArgumentOrSelectionSaysHowToSelect` asserts, and that test runs the real command through the real resolver, so the contract is pinned even though the resolver's implementation sits outside this diff.

CodeStaple:
Second review pass addressed — three fixes in cb04f94, one answered here.

- **[low] Missing total_pages treated as an exhausted listing** — fixed in cb04f94: `findApplicationByRepository` now terminates only on an empty page; a non-empty page whose pagination reports no page count is refused as undecidable ("did not report its page count") rather than answered from partial data — the wrong answer to "is this registered?" registers a duplicate application. Covered by `TestApplicationShipRefusesAnUndecidableApplicationsListing`. (The same shape exists in the pre-existing `resolveApplicationID`; left untouched there since its worst case is a wrong not-found, not a duplicate write, and it's outside this PR's lanes.)
- **[low] Supplementary github read skipped if any github credential is present** — fixed in cb04f94: the provider-filtered read now runs unconditionally on the unfiltered path, so a backend omitting only some git credentials (e.g. only the App-installation-backed ones) is still completed; the id-dedupe keeps already-listed rows single. Covered by `TestCredentialsListMergeDeduplicatesAndCompletesPartialOmission`.
- **[low] Server-side workflow-runs error degrades to "no runs yet"** — fixed in cb04f94: ten consecutive error-carrying payloads now fail the command with the server's own message (exit 1, not the wait-timeout exit), pointing at `ankra credentials repositories`; a recovered listing resets the counter. Covered by `TestApplicationShipFailsOnAPersistentWorkflowListingError`.
- **[low] Org-drift guard silently passes when OrganisationID is empty** — not changed, deliberately. The CLI check is an early courtesy, not the enforcement: every cluster read ship can make (`GetCluster`/`GetClusterByID`/`ListClusters`) is already scoped to the resolved organisation server-side, and the deploy itself re-validates the cluster against the organisation and answers cluster-not-found for a foreign id (cluster-api `deploy.go`, `DeployFailureClusterNotFound`) — a blank OrganisationID cannot let a cross-org deploy through, it can only skip the earlier, friendlier message. Failing loudly on an empty field would instead break ships driven by a `selected.json` written by an older CLI whose payload lacked the field, for clusters that are in the right organisation. The echoed organisation + cluster names before acting remain the human-facing defence.

CodeStaple:
First live production run (org Ankra - Development, cluster playground, existing app shipfortune, `--set replicas=1 --set image_tag=… --set ingress_*`) surfaced two real defects; both fixed in 3587bc0.

- **Adopting an in-flight deploy silently dropped the caller's `--set` values.** Ship followed an older replicas=0 park deploy and never said the supplied values were not applied. Now supplied `--set` values ALWAYS produce a deploy: an in-flight installation is waited to settlement first (`waitForInstallationToSettle`, resumable on timeout) and then the caller's deploy is issued, and even an already-healthy installation is deployed again to apply them — adoption without a new deploy is reserved for flag-less resumes. Tests: `TestApplicationShipWaitsOutAnInflightDeployWhenSetValuesAreSupplied`, `TestApplicationShipRedeploysAHealthyInstallationToApplySetValues`.
- **`Live:` printed for a scaled-to-zero workload.** "0 running of 0" satisfied the installation health wait vacuously and the URL read then produced a Live line for a URL serving 503. The Live line is now gated on pods actually Running with every container Ready (parsed from the `ready/total` cell); a namespace verified at zero running pods reports the deployment as **parked** (state `parked` in `-o json`, no URL) and exits non-zero with the way out (`--set replicas=1`). No HTTP probe — pod readiness is the right tier for private clusters. A failed pods read is distinguished from an empty namespace and does not fake a parked verdict. Test: `TestApplicationShipReportsAParkedWorkloadInsteadOfLive`.

PR body and CHANGELOG updated to match.

CodeStaple:
Fourth live finding, fixed in 7d8905a: **ship resolved applications by repository only, which hijacked the wrong app on a multi-application repository.** `ankra application ship <checkout>/pulsecheck --name pulsecheck` against CodeStaple/shipfortune matched the existing "shipfortune" app by repo, honestly noted `--name` was ignored — and then redeployed shipfortune with pulsecheck's ingress host, taking shipfortune's own hostname dark. The platform genuinely supports several applications per repository rooted at different paths.

`--name` is now part of the application's identity (`selectShipApplication`): a name-matched application is adopted, a repository whose applications all carry OTHER names registers the named one fresh via the add flow (announcing the existing names), and a multi-application repository without `--name` is refused with the names listed rather than one picked silently. Single-application repositories behave exactly as before, and `--name` no longer appears in the ignored-registration-flags note since it is never ignored. Tests: `TestApplicationShipWithADifferentNameRegistersASecondApplication`, `TestApplicationShipWithAMatchingNameAdoptsThatApplication`, `TestApplicationShipRefusesAMultiApplicationRepositoryWithoutAName` (single-app adoption stays pinned by `TestApplicationShipIsIdempotentAcrossARerun`).

PR #232, opened 2026-09-04
feat(security): sbom containers, export, workload scope and -o on every sbom read

CodeStaple:
## What

Completes `ankra security sbom` against the platform's new bill-of-materials reads (cluster [#2528](https://github.com/ankraio/cluster/pull/2528)), and fixes a shipped bug in the family.

**Fix — `-o json|yaml` never worked on the sbom family.** `security.go` registers the shared structured-output flag through `registerStructuredOutputFlags(...)`, but `security_sbom.go`'s `init` never called it, so every one of `namespaces`, `pods`, `sbom`, `sbom images` and `sbom image` promised `-o json` in its help and answered `unknown shorthand flag: 'o'`. Registered for all of them and the new command.

**`ankra security sbom containers`** — every container the scoped clusters run right now, init containers included, with or without a bill of materials. `ABSENT` rows first, an inventory line that counts the scope before `--status`/`--search` narrow it, `--status present|absent|any`, `--cluster`, `--namespace`, `--workload-kind`, `--workload-name`.

**`ankra security sbom export <digest or reference>`** — CycloneDX 1.5 JSON (default) or `--format csv`, written to the platform's file name in the current directory, to `--output-file`, or to stdout with `--output-file -`.

**`--workload-kind` / `--workload-name`** on `sbom` and `sbom images`.

Client: `ListSecuritySBOMContainers`, `ExportSecuritySBOMImage` (a raw authenticated download that keeps the `Content-Disposition` file name and maps 401/403 like the reads), and the workload selector on the component and image options.

## Testing

Command tests cover the flag pass-through and structured output, the containers rendering (absent-first, init marker, bare pod, inventory line, `--status` validation) and both export paths (stdout verbatim, file write). Client tests pin the query encoding for containers and the workload selector, and the export download with its file name. `go vet ./... && go test ./...` green; the pre-commit hook ran the linter at 0 issues.

Depends on the cluster PR for the endpoints; against an older API the two new commands answer the platform's 404.

https://claude.ai/code/session_013j2ViDDygzZG3cQB4v1ErN

Bead: ankra-0ugaj

### DIFF
--- a/cmd/e2e_test.go
+++ b/cmd/e2e_test.go
@@ -1002,6 +1002,14 @@ func (m baseMock) GetSecuritySBOMImage(client.SecuritySBOMImageOptions) (*client
 	return nil, errors.New("not implemented")
 }
 
+func (m baseMock) ListSecuritySBOMContainers(client.SecuritySBOMContainersOptions) (*client.SecuritySBOMContainerList, error) {
+	return nil, errors.New("not implemented")
+}
+
+func (m baseMock) ExportSecuritySBOMImage(client.SecuritySBOMExportOptions) (*client.SecuritySBOMExport, error) {
+	return nil, errors.New("not implemented")
+}
+
 func (m baseMock) CreateAPIToken(name string, expiresAt *string, scopes []string) (*client.CreateAPITokenResponse, error) {
 	return nil, errors.New("not implemented")
 }
--- a/cmd/security_sbom.go
+++ b/cmd/security_sbom.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"os"
 	"strings"
 
 	"ankra/internal/client"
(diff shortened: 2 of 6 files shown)


CodeStaple:
Addressed in b1d2848 (@ankraai):

- **Server-supplied file name used as a path** — fixed. The Content-Disposition name is reduced to its base name (`sbomExportLocalFileName`); an empty, `.`, `..` or hidden result falls back to a format-derived `sbom.cdx.json` / `sbom.csv`. `TestSecuritySbomExportNeverTreatsTheServerFileNameAsAPath` pins `../../etc/passwd` → `passwd`, `/tmp/evil.json` → `evil.json`, and an end-to-end run with a hostile name landing in the current directory.
- **`--format` not validated client-side** — fixed. `normalizeSbomExportFormat` accepts `cyclonedx` / `cdx` / `json` / `csv` and turns anything else into a usage error before any request is made.
- **Export overwrites without warning** — fixed. An existing target is refused with a hint unless `--force` is passed; the test covers refuse-then-force.
- **Possible double registration of `-o`** — no change needed. None of `namespaces`, `pods`, `sbom`, `sbom images` or `sbom image` had the flag before (that absence is the bug this PR fixes: their help promised `-o json` and cobra answered `unknown shorthand flag`), and `registerStructuredOutputFlags` guards with `Flags().Lookup("output") == nil` before adding, so a second call is a no-op rather than a panic.

CodeStaple:
Addressed in a085e19 (@ankraai): `--status` is normalised once and the canonical lower-case value is what reaches the platform (`--status " ABSENT "` → `absent`, pinned by the test); the unreachable separator check in the file-name helper is gone.

PR #233, opened 2026-09-04
feat(security): sbom findings lists the CVEs on one image (ankra-0ugaj)

CodeStaple:
Stacked on #232 (base `feat/security-sbom-complete`); retarget to `master` once that lands. Pairs with cluster #2537.

The image detail counted findings per component and stopped there; the list behind that count did not exist.

## What changes

- **`ankra security sbom findings <image>`** lists every CVE the scanner names on one image: one row per CVE and installed package version, aggregated across the containers running it, with the fixed version, the CISA KEV listing and EPSS probability, the worst disposition across its occurrences, and how many workloads and clusters run it. It answers for an image without a bill of materials too and prints **ABSENT** in the header so the gap stays visible. `--severity` (repeatable), `--search`, `--sort`, `--order`, paging, `-o json|yaml`.
- **`ankra security namespaces`** rows carry `sbom_images`; the SBOM coverage block on the inventories follows `--namespace` / `--workload-*` (backend change, no CLI code).
- Client read `ListSecuritySBOMImageFindings` on `/sbom/image/findings`; `APIClient` interface + `baseMock` extended; changelog under Unreleased.

## Verification

`go build`, `go test ./cmd/ ./internal/client/`, golangci-lint clean (pre-commit hook).

Bead: ankra-0ugaj

https://claude.ai/code/session_013j2ViDDygzZG3cQB4v1ErN

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 # Ankra CLI Changelog
 
+## Unreleased
+
+### Added
+
+- **`ankra security sbom findings <image>` lists the CVEs on one image.** The
+  image detail counted findings per component and stopped there; this is the
+  list behind the count: one row per CVE and installed package version,
+  aggregated across the containers running the image, with the fixed version
+  when one exists, the CISA KEV listing and EPSS probability, the worst
+  disposition across its occurrences, and how many workloads and clusters run
+  it. It answers for an image without a bill of materials too - the CVEs come
+  from the vulnerability reports - and says ABSENT in the header so the gap
+  stays visible. `--severity` (repeatable), `--search`, `--sort` and paging
+  narrow it; `-o json` for scripts. Pairs with `ankra security finding <id>`
+  for one CVE in full.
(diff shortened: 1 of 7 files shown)


CodeStaple:
Addressed in 3b0c9b0 (@ankraai):

- **KEV caveat on empty pages** - fixed. Both feed caveats now print above the table via `intelligenceCaveats`, before the empty-page return, and the summary's known-exploited figure reads `unknown` while `kev_synced_at` is null. The test drives an empty page and asserts the caveat survives it.
- **Unknown SBOM status** - fixed. `sbomStatusLabel` renders `present`, `ABSENT` for an explicit absent, and `unknown - this platform version does not report ...` for anything else (an older server's empty string included); pinned by the test.
- **EPSS sync state** - fixed. `epssText` already renders nothing for a missing score, so the gap was the disclosure: a second caveat now says EPSS has not been synced when `epss_synced_at` is null, so a blank probability reads unknown, not low.

PR #235, opened 2026-09-04
fix(stack-profiles): say when list is showing a page, not the catalogue (ankra-e28tp, PLA-824)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-824/1146-ai-chat-on-a-stack-profile-page-does-not-receive-which-profile-is
Bead: ankra-e28tp

## What

`ankra stack-profiles list` pages at 25 by default and printed that page as a
plain table with nothing marking it as partial. A customer organisation with 35
profiles saw 25 rows and no sign the other 10 existed:

```
$ ankra stack-profiles list | grep -c '^│ [a-z]'
25
$ ankra stack-profiles list --page-size 100 -o json | jq .total_count
35
```

The listing reads as complete, and `--page-size` is only discoverable if you
already suspect something is missing.

`StackProfileListResponse.TotalCount` was already on the response and simply
unused. The table now gets a footer whenever the page does not cover the
catalogue:

```
Showing 25 of 35 stack profiles (page 1 of 2); use --page/--page-size to see the rest.
```

Follows the existing convention in `cmd/tickets.go` ("Showing %d of %d
tickets; raise --limit to see more.").

## Scope

- Table mode only. `renderStructured` returns before the footer, so
  `-o json|yaml` stdout stays parseable and byte-identical.
- No behaviour change when the page already covers every profile.
- The page count is omitted rather than guessed if `--page-size` is
  non-positive.

## Where this came from

Reported via Linear PLA-824 (Ankra ticket #1146), whose headline complaint is
on the portal side — the AI chat's fallback picker offering 12 options while
saying there are 35. That half is **not** in this PR; it is a separate
truncation in the chat tool lane, tracked on `ankra-14ar6.11`. This PR is the
CLI half the reporter explicitly asked for ("print a 'showing N of TOTAL'
footer when more pages exist").

## Tests

`cmd/stack_profiles_list_test.go`, driving the real command through
`RunE` + `captureStdout`:

- a 25-of-35 page prints the footer, the page position, and how to see the rest
- a complete listing prints no footer and still renders the table
- the last page counts as `page 2 of 2`
- a missing page size drops the page count instead of computing a wrong one

Local: `go build ./...`, `go test -count=1 ./...` (green), `go vet ./...`
(clean), `gofmt` clean. `golangci-lint` is not installed on this machine, so
the full linter runs in CI.


Myothas:
**🤖 tech-agent** — re: Ankra AI review finding

**[medium] Zero/absent total_count is treated as 'listing is complete'** (`cmd/stack_profiles.go:152`) — **valid, fixed in 7b8e2e6.**

The finding is right, and it is the same absent-answer bug this PR exists to fix. `StackProfileListResponse.TotalCount` is a plain `int` (`internal/client/stack_profiles.go:25`), so a response that omits `total_count` unmarshals it to `0`, and the old `totalCount <= shown` guard put "the total is unknown" down the same branch as "the page already covers the total". A full 25-row page from a server that reports no total went back to reading as the whole catalogue.

Unknown is now its own answer:

- `totalCount <= 0` **and the page is full** (`shown >= pageSize`) → `Showing 25 stack profiles (page 1; catalogue size unknown); use --page/--page-size to check for more.` The count is never printed as a total.
- `totalCount <= 0` **and the page is short** → still silent. Fewer rows than the page size means the server had no more to give, so there is nothing the reader is missing.
- Known totals are unchanged.

Two tests added: `TestStackProfilesListSaysSoWhenTheTotalIsMissing` drives the real command with `total_count` absent and asserts the unknown-size wording (and that ` of 0 ` never appears), and `TestStackProfileListingFooterStaysQuietOnAShortPageWithoutATotal` pins the short-page case.

Local on 7b8e2e6: `go build ./...`, `go vet ./...` clean, `go test -count=1 ./...` green, `gofmt` clean on both changed files.

Myothas:
**🤖 tech-agent** — re: Ankra AI review on `7b8e2e6`

**[medium] Confirm `-o json`/`yaml` returns before the footer print** — **confirmed, and now pinned by a test in `5bce732`.**

The early return is real, it just wasn't in the diff. `cmd/stack_profiles.go:118-120`:

```go
if rendered, err := renderStructured(cmd, response); rendered || err != nil {
    return err
}
```

and `cmd/output.go:36-45` returns `false` only for `outputDefault` (no `-o`), `true` for json and yaml:

```go
if format == outputDefault {
    return false, nil
}
return true, encodeStructured(cmd.OutOrStdout(), format, value)
```

So with `-o json|yaml` the command returns ~20 lines above the footer print, and the table branch is never reached.

You were right that a comment is not evidence, so I have replaced it with one: `TestStackProfilesListKeepsStructuredOutputParseable` runs the real command with `-o json` and requires the *whole* of stdout to `json.Unmarshal` into one document, with no `Showing` anywhere in it. If a later edit ever moves the footer above that early return, this fails instead of silently breaking every caller piping into a parser.

(One wrinkle worth recording: the test resets `rootCmd`'s writers first. `executeCommand` in `testutil_test.go:16-23` points them at its own buffer and never restores them, and cobra's `OutOrStdout` walks to the parent — so without the reset the JSON lands in whichever earlier test ran, not on the stdout `captureStdout` is watching.)

---

**[low] Footer suppressed when total equals rows shown on a later page** — **no change; the branch is unreachable for any self-consistent response, and the suggested guard trades one wrong output for another.**

The condition needs `totalCount <= shown` while `page > 1`. Serving page `p` means an offset of `(p-1)*pageSize`, so a correct server returns at most `totalCount - (p-1)*pageSize` rows. For `p >= 2` that gives `shown <= totalCount - pageSize < totalCount`, which makes `totalCount <= shown` impossible unless the response contradicts itself — reporting a total smaller than the offset it just served past.

Your concrete example (`total=25, page=2, pageSize=25`) is that contradiction: 25 rows served from row 26 onward, of a catalogue of 25. And it is also where the proposed fix goes wrong — `page > 1` would fall through to the footer and print `(page 2 of 1)`, since `totalPages` computes to 1 from the same inconsistent total. Swapping a suppressed footer for a self-contradicting one is not an improvement.

`page*pageSize` against `totalCount` is not the intent either: on the honest last page (`shown=10, total=35, page=2, pageSize=25`) it reads as "the reader has now seen all 35" and suppresses the footer — but the table in front of them holds 10 rows out of 35, which is exactly the case this PR exists to caption. `TestStackProfileListingFooterCountsTheLastPage` pins that.

Happy to be overruled by a human reviewer if you'd rather have defensive handling for a self-inconsistent API response, but I don't think it earns its complexity here.

---

Local on `5bce732`: `go build ./...`, `go vet ./...` clean, `go test -count=1 ./...` green, `gofmt` clean on both changed files.

Myothas:
**🤖 tech-agent** — re: Ankra AI review on `5bce732`

**[low] Test assumes `--output` is a local flag on `stackProfilesListCmd`** (`cmd/stack_profiles_list_test.go:55`) — **checked; it is local, so no change. You were right to flag it as unconfirmable from the diff.**

The registration is one hop outside the diff, which is why it did not show. `cmd/stack_profiles.go:637`:

```go
registerStructuredOutputFlags(stackProfilesListCmd)
```

and `cmd/output.go:9-15` puts the flag on the command's **own** flag set, not on a parent:

```go
func registerStructuredOutputFlags(cmds ...*cobra.Command) {
	for _, command := range cmds {
		if command.Flags().Lookup("output") == nil {
			command.Flags().StringP("output", "o", "", "Output format: json or yaml (default: human-readable)")
		}
	}
}
```

Each command that wants structured output opts in individually (`stackProfilesListCmd`, `stackProfilesGetCmd`, `stackProfilesApplyCmd` at `cmd/stack_profiles.go:637-656`). `rootCmd`'s persistent flags are only `config`, `token`, `base-url` and `org` (`cmd/root.go:119-126`) — `output` is not among them, so there is no persistent definition for `Flags().Set` to miss.

Empirically too: the whole `cmd` package is green locally on `5bce732`, all 1403 tests including the four that go through `runProfilesListAs`. A `no such flag` would have failed them at the `t.Fatalf` you pointed at, not passed quietly.

The pattern you were guarding against is real for this codebase — `--org`/`--token` *are* persistent on `rootCmd`, and a test setting one of those through `stackProfilesListCmd.Flags().Set` would break exactly as you describe. It just is not the case for `--output`.

---

**CI status on this PR is unrelated to the change.** Both runs on `5bce732` are still inside the `Run tests` step after ~1h45m. That is the runner, not the suite: the one run that completed on this branch spent 17m in `golangci-lint` and 19m in `Run tests` and passed, and each push here fires two duplicate jobs (`push` + `pull_request`) that compete for the same pool. Locally, the exact CI command —
`go test -race -count=1 -timeout=120s -coverprofile=coverage.out ./...` — is green on `5bce732`, every package well inside the 120s per-binary timeout:

```
ok  ankra/cmd                     43.485s  coverage: 74.1% of statements
ok  ankra/internal/client         14.876s  coverage: 58.5% of statements
ok  ankra/internal/kubeconfig      2.152s  coverage: 80.9% of statements
ok  ankra/internal/migrate         6.785s  coverage: 83.4% of statements
ok  ankra/internal/migrate/docker  3.548s  coverage: 84.8% of statements
ok  ankra/internal/skills          4.012s  coverage: 85.9% of statements
```

`go build ./...` and `go vet ./...` clean, `gofmt` clean.

Myothas:

_Updated 2026-09-05 17:37Z (original escalation 2026-09-04 23:51Z; this comment is edited in place rather than re-posted)._

**PR:** https://github.com/ankraio/ankra-cli/pull/235 — `ankra stack-profiles list` prints a `Showing 25 of 35 stack profiles (page 1 of 2); use --page/--page-size to see the rest.` footer whenever the page does not cover the catalogue, so page 1 stops reading as the whole set. Table mode only; `-o json`/`yaml` stdout is untouched. 3 files, +239/-0, tests included.

**State:** `ankra-cli#235  OPEN  ready-for-review since 2026-09-04T21:26:01Z  ci=PENDING  review=none  ai-review=NEUTRAL  merge=UNSTABLE  (read 2026-09-05T17:36:42Z)`

**What moved since the original escalation:**

- 2026-09-05 16:37Z — `origin/master` was merged into the branch (`9dcbcd0`, includes the v0.15.0-rc2 CHANGELOG move). Thank you. I did not make that commit, but I verified it: `go build`, `go vet` and the exact CI test command (`go test -race -count=1 -timeout=120s ./...`) are green locally on `9dcbcd0`.
- 16:39Z — both `lint-test-build` runs started on `9dcbcd0`. The `pull_request` run (`33978545732`) lost its runner inside `golangci-lint` (step frozen `in_progress`, later steps never started, no log uploaded) and was marked `failure` at 17:18Z. The `push` run (`33978544140`) passed lint in 11m and has been in `Run tests` since 16:50Z.
- 17:28Z — someone re-ran the failed `pull_request` run (attempt 2); it is now past lint and in `Run tests`.
- 17:36Z — I pushed an **empty commit** `5bf3e18` (no source change) so the head carries fresh check-runs instead of a dead one; a check-run belongs to its commit and cannot be re-run from my session. Both new runs (`33981461675`, `33981465355`) are in progress.

**Why I did not merge it:** merge-gate condition 2 — *every required check is green*. Nothing on the head has concluded yet. Every other condition holds: my own PR, not scary-list, 3 files / +239, merges cleanly, AI review resolved (four findings over three rounds, all fixed or answered above; the latest run on this head concluded NEUTRAL with no new findings), no human has requested changes, and PLA-824 passed the injection screen.

1. **If the runs on `5bf3e18` come back green, this is a plain merge** — squash-merge it, or leave it and my next pass will merge it on the gate. If they wedge again, that is the runner pool (`ankra-1h1kp`: two duplicate jobs per push, no concurrency group, jobs lost mid-step), not the change.
2. **The one judgement call in the diff:** `stackProfileListingFooter` treats a `total_count` of `0` as *unknown* rather than *empty*, because the field is a plain `int` and an omitted `total_count` unmarshals to `0`. A full page with no reported total prints `catalogue size unknown` instead of staying silent. I believe that is right — the whole bug was a page passing for the catalogue — but it is the one place the output changes for a response shape I could not observe from a real server.

Myothas:
**🤖 pr-shepherd** — blocked on the runner outage, not on this change

Picking up where the escalation above left off. The runs on `5bf3e18` did **not** come back green, so the "if they come back green, this is a plain merge" branch never opened.

**What I did:** re-ran the dead `push` job as attempt 2 (`33981461675`). It got further than any run since the outage began — `Verify embedded skills are in sync` ✅ and `Run golangci-lint` ✅, both at normal speed — then sat in `Run tests` for 77 minutes (12:05:45 → 13:22:34Z) before GitHub killed it with the same annotation as every other run:

> The self-hosted runner lost communication with the server.

**This is not specific to this PR.** `lint-build.yml` has not produced a single success on *any* branch since **2026-09-05T15:00Z** — roughly 22 hours. `master`, `ankra-htxvm.3-cluster-routes-check`, `feat/ankra-vevro-cli-org-ci-settings` and this branch all fail the same way. Tracked as `ankra-k7vll`, whose read (the `-race` build starving the 2-vCPU Depot runner's memory) matches exactly what I see: the failure is isolated to the `-race` test step and never touches lint.

One measurement worth recording for anyone tempted to wait it out rather than retry: the single run that *did* succeed on this branch (`33923013121`) took **3h44m**, with `Run tests` alone at **3h16m** for a suite that finishes in ~60s locally. So a long-running job here is starving rather than hung — but as a strategy, retrying is a poor lottery.

**State of the change itself** — everything except the test step is verified:

| Gate | Status |
| --- | --- |
| Conflicts | ✅ MERGEABLE; the `origin/master` merge put the Fixed entry under `## Unreleased`, above the new `v0.15.0-rc2` heading |
| Un-drafted | ✅ ready since 2026-09-04 |
| AI review | ✅ verdict *looks good*; 4 findings over 3 rounds, 2 fixed and 2 answered (both `[low]`) |
| Lint / skills-sync | ✅ passed **in CI** on this head |
| Tests | ⛔ never concluded in CI; green locally on the exact CI command across three independent runs |
| Scary list | ✅ none apply — CLI output text plus tests, 3 files, +239/−0 |

I spot-checked the two findings that were answered rather than fixed, instead of taking the replies on trust. Both hold: `registerStructuredOutputFlags` puts `--output` on the command's own flag set (`cmd/output.go:9-15`), so the test helper resolves it; and reading an absent `total_count` as *unknown* rather than *empty* is right, since a genuinely empty catalogue still stays silent through the `shown >= pageSize` guard.

**I am not merging it.** The merge gate wants a green suite on the head commit, and this suite has never run to completion there. Strong local evidence is not the same thing, and `master` is unprotected, so a merge here is immediate with nothing to catch a mistake.

**Unblocking path:** #244 is the critical path for every open ankra-cli PR, and it has had no run at all since its own was cancelled on 09-05T13:01 — the ARC path is still unvalidated. Once it lands, this PR needs only a re-run, no new commits. I have deliberately left #244 alone: it is a draft on another lane and touches the codesigned darwin release legs.

PR #239, opened 2026-09-05
fix(playground): status, destroy and resize take the cluster's name as well as its id (ankra-y8l44.35)

Myothas:
## What & why

Seen on 2026-09-02 (demo org): `ankra cluster playground status playground` and `ankra cluster playground destroy playground` answered a bare **404 Not Found**. The playground routes want the cluster id; the CLI passed the name through untouched, and a non-UUID path segment gets the same 404 body as a missing environment, so nothing said an id was expected.

- `resolvePlaygroundClusterID`: an id-shaped argument (`isLikelyClusterID`) goes straight through; anything else resolves through the existing `resolveClusterID` walk of `cluster list` (case-insensitive name match); an unknown name is refused **before any request** with the list command named.
- Applied to `status`, `destroy` and `resize`; their `Use` strings read `<cluster_id|name>`.

Bead: ankra-y8l44.35 (parent ankra-y8l44; the platform side of the same incident is cluster#2655).

## Test plan

- `TestPlaygroundCommandsResolveAClusterName` (status, destroy and resize by name reach the API with the id), `TestPlaygroundCommandsNameAnUnknownClusterInsteadOfA404`.
- Existing playground tests moved to an id-shaped fixture id (they used `cluster-42`, which now reads as a name).
- `go test ./...` green; pre-commit (go test + golangci-lint) passed.

### DIFF
--- a/cmd/cluster_playground.go
+++ b/cmd/cluster_playground.go
@@ -79,7 +79,7 @@ var clusterPlaygroundPlansCmd = &cobra.Command{
 var clusterPlaygroundResizeSize string
 
 var clusterPlaygroundResizeCmd = &cobra.Command{
-	Use:   "resize <cluster_id>",
+	Use:   "resize <cluster_id|name>",
 	Short: "Resize a paid playground to another size",
 	Long: "Change a paid playground to another paid size in place: only the namespace quota and " +
 		"the monthly price change (the month is billed pro-rata across the change), nothing you " +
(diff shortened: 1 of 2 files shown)


Myothas:
CI note (2026-09-05 11:30Z): `lint-test-build` sits in **Run tests** for hours on this PR, and so does the master run started at 10:59Z that does not contain this change. The step normally takes ~20 s and `go test` runs with `-timeout=120s`, so a hung test would have failed within two minutes; this is the runner, not the suite. Locally: `go test -race -count=1 ./...` green in 2 s, pre-commit golangci-lint green. Cancelled and re-ran once (same stall). Leaving the PR for a green run once the runners recover.

PR #241, opened 2026-09-05
feat(stack-profiles): import --as-draft opens the document as a builder draft (ankra-ry5p3)

Myothas:
## What

`ankra stack-profiles import <file> --as-draft` (Smartoptics #1093, PLA-797): stage a file-authored profile candidate for human review instead of publishing a version immediately.

- New flag `--as-draft` on `stack-profiles import`. The document opens as a builder draft on the named profile: it attaches to your organisation's existing profile of the same name as the next-version candidate when one exists (the command reports the base version), and opens a new-profile draft otherwise. Nothing is published; the command prints the draft id and the `drafts validate` / `drafts publish` follow-ups.
- New client method `ImportStackProfileAsDraft` posting to `POST /api/v1/org/stack-profiles/import-draft` (ships in ankraio/cluster#2664; this PR should merge after that one so the flag has a live endpoint to talk to).
- `--category` is only sent when explicitly set: a draft attaching to an existing profile applies non-null metadata at publish, so sending the flag default (`general`) would overwrite the profile's own category.
- Import warnings (redacted secrets, dropped parameter definitions, bare `${var}` findings) are printed before the draft summary.

Command docs regenerate from the updated `Long` text via `tools/gendocs` on the next release tag.

## Tests

- `TestImportStackProfileAsDraft_PostsAndDecodes`: method/path/payload shape (unset metadata omitted from the body), draft + warnings decode.
- `TestImportStackProfileAsDraft_SurfacesBackendDetail`: the open-draft 409 detail surfaces as the command error.
- Full `go build ./...` + `go test ./...` + lint green via the pre-commit hook.

Bead: ankra-ry5p3


Myothas:
Answering the two findings:

- **[low] Draft field availability assumed from another file** - confirmed against the declaration: `StackProfileDraft` in `internal/client/stack_profile_drafts.go` carries `ProfileID *string`, `BaseVersion *int` and `Parameters []map[string]any`, so the dereferences and the test's parameters read match. The tree compiles (the push-event `lint-test-build` run on this commit is green; the pull_request-event twin failed with its log evicted and passed on re-run).
- **[low] Existing-profile message prints draft name, not profile identity** - fixed in dacf06f: the attach message now prints the profile id alongside the name, so the target profile is unambiguous even where the names differ only by case (the server resolves the attach case-insensitively, so the draft's name and the profile's name can differ in casing but never in identity).

PR #244, opened 2026-09-05
ci: run on the organisation's self-hosted ARC scale set instead of Depot (ankra-k7vll)

Myothas:
## What & why

On 2026-09-05 every ankra-cli job on `depot-ubuntu-24.04` ran 20-60x slower than usual: `golangci-lint` 19 min for a 17 s step, `go test` 20 min for a 23 s step, and a master run died after 46 min with the tests step never concluding. Other repositories on the same Depot label (ankra-docs, cluster PRs) were normal at the same time, so this was per-repository starvation on Depot's side with nothing to pull on ours.

Neither repository has been on GitHub-hosted runners since August (ankra-cli#121 / ankra-docs#162, ankra-lo47s - now closed as stale; GitHub billing is restored, the v0.15.0-rc1 darwin legs ran on `macos-latest` on 2026-09-03). This PR moves ankra-cli's CI onto Ankra's own runners.

- **`runs-on: ${{ vars.CLI_RUNNER || 'arc-runner-set' }}`** on every Linux job. `arc-runner-set` is the ARC scale set registered for `github.com/ankraio` (runner group `arc-privileged-build`, max 6 runners, DinD); the group has been granted to this repository. Setting the repository variable `CLI_RUNNER` moves the jobs elsewhere without a code change.
- **Go jobs run in `golang:1.26`** (lint-test-build, docs-sync, systemtest, create-release, publish-homebrew), the same pattern as cluster's ARC jobs: the bare runner image has no make, gcc (the race detector needs cgo) or Go.
- **`.github/scripts/install-tooling.sh`** installs `gh` from GitHub's apt repository plus any extra packages (the homebrew job needs an ssh client), as root in the container or via sudo on the bare runner; a no-op where gh is preinstalled (Depot). Verified inside `golang:1.26`.
- **Release matrix**: the linux/windows legs move to the knob without a container (the job is shared with the macOS legs); `gh` is installed on Linux legs only. The **darwin legs stay on `macos-latest`**: ad-hoc `codesign` is macOS-only.
- **`timeout-minutes`** on every job (20/15/30), so a starved runner fails in minutes instead of holding a job for six hours.
- **lint-build runs once per commit**: push on `master`, `pull_request` elsewhere, with superseded branch runs cancelled. It used to run twice per PR commit, which matters on a six-runner pool.

Not in this PR: ankra-docs. It is healthy on Depot, and its jobs need Python, Node, buildx/QEMU and pip-installed scanners the minimal ARC image lacks, so moving it is a separate containerising change.

Bead: ankra-k7vll

## Test plan

- This PR's own `lint-test-build` is the proof: it must run on an `arc-runner-set` runner (check the job's runner name) and finish in the usual ~1-2 min.
- `actionlint` clean on all four workflows; `bash -n` on the script; the script run end-to-end in `golang:1.26` (gh, ssh, git, make, gcc present afterwards).
- Release path is exercised by the next tag; the darwin legs are unchanged.

### DIFF
--- a/.github/scripts/install-tooling.sh
+++ b/.github/scripts/install-tooling.sh
@@ -1,0 +1,29 @@
+#!/usr/bin/env bash
+# Installs the GitHub CLI (from GitHub's own apt repository, so the version does
+# not depend on the distribution's package) plus any extra apt packages named
+# as arguments, on whichever host the job landed on: as root inside a job
+# container (golang:1.26) or through sudo on the bare self-hosted runner image.
+# Depot's images ship gh preinstalled; there this is a no-op.
+set -euo pipefail
+
+if [ "$(id -u)" -eq 0 ]; then
+    SUDO=""
+else
+    SUDO="sudo"
+fi
+
+if ! command -v gh >/dev/null 2>&1; then
+    ${SUDO} mkdir -p /etc/apt/keyrings
+    curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+        | ${SUDO} tee /etc/apt/keyrings/githubcli-archive-keyring.gpg >/dev/null
+    ${SUDO} chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+        | ${SUDO} tee /etc/apt/sources.list.d/github-cli.list >/dev/null
+    set -- gh "$@"
+fi
+
+if [ "$#" -gt 0 ]; then
+    ${SUDO} apt-get update -qq >/dev/null
+    DEBIAN_FRONTEND=noninteractive ${SUDO} apt-get install -y -qq --no-install-recommends "$@" >/dev/null
+fi
+gh --version | head -n 1
--- a/.github/workflows/docs-sync.yml
+++ b/.github/workflows/docs-sync.yml
@@ -20,13 +20,19 @@ permissions:
 jobs:
   sync-docs:
     name: Generate and PR CLI reference
-    runs-on: ${{ vars.CI_RUNNER_2 || 'depot-ubuntu-24.04' }}
+    runs-on: ${{ vars.CLI_RUNNER || 'arc-runner-set' }}
+    container:
+      image: golang:1.26
+    timeout-minutes: 20
     steps:
       - name: Checkout ankra-cli
         uses: actions/checkout@v6
         with:
           # Full history so manual runs can name a version via git describe.
           fetch-depth: 0
+
+      - name: Install tooling (gh)
+        run: .github/scripts/install-tooling.sh
 
       - name: Set up Go
         uses: actions/setup-go@v5
--- a/.github/workflows/lint-build.yml
+++ b/.github/workflows/lint-build.yml
@@ -3,17 +3,35 @@ name: Lint, Test and Build
 on:
   push:
     branches:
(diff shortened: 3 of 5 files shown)


Myothas:
Parked as draft: the lint-test-build job sat in `queued` with label `arc-runner-set` and no runner for 15 min. ankra-cli is a **public** repository and the `arc-privileged-build` runner group has `allows_public_repositories=false`, so ARC will never accept its jobs as-is. Flipping that flag is a security-posture decision (fork pull requests can rewrite the workflow file, so any fork PR whose run is approved would execute on privileged DinD runners inside Ankra's infrastructure); it is being put to the repository owner before anything changes. The workflow changes themselves are complete and actionlint-clean; the runner knob defaults can be re-pointed in one line once the decision is made.

Myothas:
## Closing — the ARC scale sets are not available to this repo, and the incident that motivated this is fixed

Closing as no longer relevant. Three independent reasons:

**1. The destination is closed to ankra-cli.** The repo is **public** (`gh repo view --json visibility` → `PUBLIC`), and the `arc-privileged-build` group forbids public repos by design so a fork PR can never reach privileged dind on the cluster holding the Harbor and ArgoCD credentials — stated in #290. Public repos are in neither runner group.

This is confirmed empirically by this PR's own and only CI run: run `33967700694` requested label `arc-runner-set`, was **never assigned a runner** (`runner_name: ""`, `steps: []`, zero steps executed) and sat queued for 16m14s before a human cancelled it on 2026-09-05. Nothing has run on the branch since. That is also the failure mode if it merged — a job on an unclaimable label queues forever rather than erroring, so a `v0.16.0` tag would leave `create-release`, all four cross-compile matrix legs and `publish-homebrew` hanging indefinitely with nothing to say why. And because the PR also swaps `vars.CI_RUNNER_2` for a repo-local `CLI_RUNNER`, the escape hatch becomes a variable that does not exist yet, leaving no non-ARC lane at all.

**2. The knob this was meant to provide already existed, three days before this PR opened.** `CI_RUNNER_2/4/8` landed as org variables in `5249f58` (#221, 2026-09-02) and master's four workflows already read `vars.CI_RUNNER_2` — so "move the runners without a code change" is solved, org-wide. This PR moves away from that convention rather than toward it.

**3. The 2026-09-05 starvation incident is resolved, and by a different fix.** #248 (2026-09-04) addressed the actual root cause the bead comments identified — the `-race` compile starving a 2-vCPU runner — with `GOMEMLIMIT: 5GiB`, and master's `lint-build.yml` already carries `concurrency`/`cancel-in-progress` and `timeout-minutes: 45`. Master run `34550657110` (2026-09-11 01:26Z) finished `lint-test-build` in **73 seconds** on `depot-ubuntu-24.04`; #275 and #271 both have green 1-2 minute runs.

Worth noting the conflict is substantive, not textual: master's `lint-build.yml` has since gained `concurrency`, `timeout-minutes: 45`, `actions/setup-go@v6`, `make verify-skills` and that 25-line measurement-backed `GOMEMLIMIT` block tuned to "this runner's 8GB". Resolving in this PR's favour would silently drop GOMEMLIMIT and reinstate the very OOM/thrash that caused the outage.

**The live successor is #290** ("ci: replace the dead Depot fallback defaults", ankra-ai93r), which maps the fallback to `ubuntu-latest` rather than ARC — the right answer for a public repo.

**One idea worth salvaging:** this PR added `timeout-minutes` to every job, and `docs-sync.yml` plus both `release.yml` jobs still lack one on master. Cheap to cherry-pick onto #290 or a small follow-up.

Bead ankra-k7vll should be closed or re-scoped to "public repos keep a hosted lane" rather than left pointing at this branch.

PR #247, opened 2026-09-07
feat(security): licence risk tiers and --license-risk on the SBOM commands (ankra-7olqe)

CodeStaple:
## What

`ankra security sbom` and `ankra security sbom image` show each component's **licence risk tier** and take `--license-risk` (repeatable: `network_copyleft`, `source_available`, `copyleft`, `weak_copyleft`, `permissive`, `unknown`); `ankra security sbom images` and the image detail summarise each image's obliging tiers ("2 network copyleft, 1 copyleft", red for the two that reach the hosting service, yellow for plain copyleft).

Bead: ankra-7olqe

Pairs with the cluster PR of the same branch name (`license_risk`, `license_exposure` and the `license_risk` query parameter on the SBOM reads).

### DIFF
--- a/cmd/security_sbom.go
+++ b/cmd/security_sbom.go
@@ -552,9 +552,13 @@ func componentFindingsCell(component client.SecuritySBOMComponent) string {
 }
 
 // licenseRiskCell names the tier, red for the two that reach the code
-// hosting the package and yellow for plain copyleft.
+// hosting the package and yellow for plain copyleft. A server that sent
+// no tier at all renders a dash: only the literal unknown tier says the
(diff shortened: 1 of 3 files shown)


CodeStaple:
Thanks @ankraai - addressed in 20e274e: an empty `license_risk` renders as `-` and only the literal `unknown` tier prints `unknown`; `LicenseExposure` is a pointer, so an image the server reported no exposure for prints `not reported` in the images table and `Licences:    not reported` on the detail, while an empty exposure still reads as clean. Both cases are unit-tested.

PR #259, opened 2026-09-07
feat(chat): show tool activity in the session output (ankra-14ar6.12)

CodeStaple:
Bead: ankra-14ar6.12

`ankra chat` rendered content, status, notices, proposals and errors, and dropped the `tool_start` / `tool_result` frames, so a long turn read as silent gaps between status lines and a failed tool never showed why (review F9, ankra-14ar6). Each tool call now prints one bracketed line when it starts (`[tool list_node_groups ...]`) and one when it settles (`[tool list_node_groups ok]`, or `[tool get_pods failed (transient): Agent timeout while fetching pods.]` carrying the platform's error class). Test: `TestRenderChatTurnShowsToolActivity`.

### DIFF
--- a/cmd/chat_session.go
+++ b/cmd/chat_session.go
@@ -276,6 +276,14 @@ func renderChatTurn(events <-chan client.ChatStreamEvent, out io.Writer, errOut 
 					printLine("note: " + message)
 				}
 			}
+		case "tool_start":
+			if line := chatToolStartText(event.Data); line != "" {
+				printLine(line)
+			}
+		case "tool_result":
+			if line := chatToolResultText(event.Data); line != "" {
+				printLine(line)
+			}
 		case "action_proposal":
 			proposal, decodeError := decodeActionProposal(event.Data)
 			if decodeError != nil {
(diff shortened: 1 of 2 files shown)


CodeStaple:
3f73511 on the review:

- **[medium] Absent `success` rendered as a failure**: a `tool_result` frame with no boolean `success` now renders `[tool <name> settled]` - an absent answer is not a failed one. Test extended with such a frame.
- **[low] Test relies on `printLine` wrapping and the `strings` import**: both are in the file - `printLine` is the existing bracketed-line writer the status and notice cases use, and `strings` was already imported by the test file; CI runs the package green.

CodeStaple:
On the 3f735110 review:

- **[medium] `strings` import**: it is already imported by `cmd/chat_session_test.go` (line 8 of the import block, used by the existing tests); the package compiles and the CI test job is green on this head.
- **[low] Raw tool error text on the terminal**: intended and consistent with the existing `error` case, which already prints the server's message verbatim. Tool error strings are model-facing text the platform composes and passes through its result redaction (`applyResultRedaction`, secret detector) before they leave cluster-api, so nothing lands here that the chat transcript does not already show.

PR #261, opened 2026-09-07
fix(org): ci-settings shows every stored setting and exits 7 on the admin gate (ankra-c1lop, PLA-825)

CodeStaple:
Linear: https://linear.app/ankra/issue/PLA-825/1147-cli-has-no-way-to-discover-which-container-image-registry-an-org
Beads: ankra-c1lop
Follows: #236 (merged as c5f42fb0b while this review was in flight; this is the review's findings, rebased onto master)

## Why a follow-up

#236 added `ankra org ci-settings get|set`. Reviewing it before merge turned up three things
that should not ship in a command whose whole purpose is discoverability, and the auto-merge
sweep landed the PR before the fixes could join it. They are here, unchanged, on top of master.

## 1. `get` dropped three of the eleven stored settings

The client struct carried eight fields. The platform stores eleven: `ci_run_retention_days`,
`ci_ignore_unfixed` and `ci_egress_allowed_cidrs` were absent, so `get` never showed them and
`-o json` was a silently incomplete record — the exact gap the command exists to close, one
layer down. All three are carried now. `get` renders them (`Run retention`, `Egress allowed
CIDRs`, `Ignore unfixed findings`); `set` gains `--run-retention-days`, `--egress-allowed-cidr`
(repeatable, replaces the list, empty clears it — same contract as the image policy) and
`--ignore-unfixed`, read through `Changed` so `=false` is a write and an untouched flag is not.

## 2. The admin gate exited 6 ("re-login") instead of 7 (RBAC)

A member changing the settings hits the endpoint's admin gate. `cmd/exitcodes.go` reserves
exit 7 for exactly this — authenticated, role falls short, "ask an admin" — and maps a bare
403 to exit 6, which tells the caller to log in again. The 403 now surfaces as
`PermissionDeniedError`. That type gains an optional `Detail` for gates that write
`{"detail": "..."}` rather than the RBAC `permission_denied` shape, so the platform's own
sentence is relayed verbatim; the RBAC shape keeps its existing handling.

## 3. The client had no tests of its own

`internal/client/org_ci_settings_test.go` is new: a lane-parity table pins method, path and
the presence-only PUT body (including the explicit null that clears the cluster); one test
decodes the full production payload so a dropped field fails again; both 403 shapes map to
`PermissionDeniedError`; both refusal-detail shapes (single sentence, per-member list) are
relayed as words.

Also: the platform-builders caveat `get` prints is cut from five lines to three, and the
command help now says `--org` scopes both verbs to another organisation you administer.

## Validation

- Pre-commit gate on the commit: full `go test ./...` and `golangci-lint run` — pass, 0 issues
- `go vet ./cmd/ ./internal/client/` — clean
- `go test -race -count=1 ./cmd/ ./internal/client/` — pass
- CI on the same commit, run 34147973209 — green

## Related

- **ankra-8atb4** (closed, cluster#2722): the platform's frozen `infra_error` sentence no longer
  names the fallback setting when only the `platform_builds` flag is dark.
- **ankra-vn0bd.3.10.2** (open): expose that capability read-only in `GET /org/ci-settings`,
  so the caveat can become a printed field.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,13 +11,21 @@
   the pipeline cluster and the build fallback - were discoverable only by
   asking Ankra: a run that concluded `infra_error` naming the build fallback
   left no command that could show what the fallback was set to. `get` prints
-  all eight settings, says what an unchosen pipeline cluster costs, calls out
-  a chosen cluster that has since been deleted, and reports when every value
-  is still Ankra's own default. `set` writes only the flags you pass, so
-  raising one number cannot clear the image policy; `--cluster` takes a name
-  or an id and an empty value clears it, and `--allowed-image-prefix` is
-  repeatable and replaces the whole policy list. Reading needs organisation
-  membership, changing needs organisation admin.
+  every setting the platform stores - the pipeline cluster, the build
+  fallback, both parallelism limits, the image policy, the private egress
+  allow-list, the artifact, cache and run-history retentions, the image gate
+  and its unfixed-findings floor - says what an unchosen pipeline cluster
+  costs, calls out a chosen cluster that has since been deleted, and reports
+  when every value is still Ankra's own default; `-o json` is the whole
+  record. `set` writes only the flags you pass, so raising one number cannot
+  clear the image policy; `--cluster` takes a name or an id and an empty
+  value clears it, `--allowed-image-prefix` and `--egress-allowed-cidr` are
+  repeatable and replace their whole list, and `--ignore-unfixed=false` is a
+  write while an untouched flag is not. Reading needs organisation
+  membership, changing needs organisation admin: a member's attempt is
+  refused with the platform's own sentence and exit code 7, the RBAC code,
+  rather than the re-login code. Both act on the selected organisation, or on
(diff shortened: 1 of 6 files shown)


CodeStaple:
Review findings:

- **[medium] `yesNo` must exist** — it does: `func yesNo` is defined in `cmd/ai.go`, the same `cmd` package `org_ci_settings.go` lives in, and renders true/false as yes/no; both `lint-test-build` jobs compiled and passed on this head.
- **[low] 403 on the read path reads as an admin-gate refusal** — the order in `doCISettingsRequest` already keeps the two apart: a 403 is first handed to `PermissionDeniedFromResponse`, which recognises the platform's standard `{"detail": "permission_denied"}` shape that a non-member GET receives; only a body that is not that shape (the gate writes its own sentence) is relayed verbatim through `ciSettingsRefusalDetail`. Either way the exit code is 7, which is right for both: the caller is authenticated and their role falls short. Keying the wording off the method would only matter if the platform ever wrote the gate sentence on a GET, and the GET has no gate.

PR #263, opened 2026-09-07
fix(application): add judges the GitHub App per repository, not per credential (ankra-ctsmd)

CodeStaple:
## What this fixes

Walking the CI/CD flow as a first-time user, with a brand-new repository, the first command refused:

```
$ ankra application add .
Error: no available GitHub credential found; install the Ankra GitHub App, then run this command again
```

The App was installed and working. What had happened is that one application in the organisation is bound to a repository in a *different* GitHub account, which that installation can never reach. Repository coverage is one verdict for the whole credential, so the credential latches `down`, `available` goes false, and this command — which filtered candidates on `available` — saw no credential at all. The message then sends the user to reinstall an App that is already there, and no amount of reinstalling fixes it.

## What changed

Availability is a verdict over every repository every application binds to a credential. The question `application add` actually has is narrower and belongs per repository: *can this installation read the repository being added*.

- A credential is a candidate whether or not it is globally available; availability only ranks candidates, so a healthy credential still wins a tie between two for the same owner, and an explicitly named one is no longer refused for a verdict about somebody else's repository.
- After selection, the repository is checked against the coverage the platform already publishes (`GET /org/credentials/{id}/repositories`). A repository the installation genuinely cannot see is refused by name with the page that grants it and the setting that stops it recurring:

```
Error: the Ankra GitHub App installation behind credential "github-app-codestaple" cannot see CodeStaple/ankra-simplest
Grant it access and run this command again, or select every repository so a new one needs no such step:
  https://github.com/settings/installations/145425934
```

- Only a **complete** listing is a negative answer. A listing GitHub truncated or refused says nothing about this repository, and is not turned into "your repository does not exist".
- The settings URL follows the account type: a personal account's installations live under the user's settings, an organisation's under the organisation's, and an installation with no stored id gets the account's list rather than a wrong deep link.

## Tests

`TestSelectApplicationCredentialKeepsAnUnavailableCredential`, `TestSelectApplicationCredentialPrefersTheHealthyOwnerMatch`, `TestInstallationSettingsURL`, and the existing `TestSelectApplicationCredential` updated to the new contract. Full `go test ./cmd/ ./internal/...` and `golangci-lint` are green (the pre-commit hook ran both).

Verified live against the Ankra - Development organisation: the message above is the real output of the fixed binary on a new repository, where the old binary printed the misleading one.

Bead: ankra-ctsmd

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -234,6 +234,10 @@ func resolveApplicationAddPlan(command *cobra.Command, repositoryPath string) (a
 	if selectionError != nil {
 		return applicationAddPlan{}, selectionError
 	}
+	if reachabilityError := credentialReachesRepository(
+		selectedCredential, repository.Owner, repository.Name); reachabilityError != nil {
+		return applicationAddPlan{}, reachabilityError
+	}
 
 	return applicationAddPlan{
 		repository:    repository,
@@ -584,12 +588,6 @@ func selectApplicationCredential(
 	if requestedCredential != "" {
 		for _, credential := range githubCredentials {
 			if credential.ID == requestedCredential {
-				if !applicationCredentialAvailable(credential) {
-					return client.Credential{}, fmt.Errorf(
-						"GitHub credential %q is not available",
-						credential.Name,
-					)
-				}
 				return credential, nil
 			}
 		}
@@ -609,12 +607,6 @@ func selectApplicationCredential(
 				),
 			)
 		case 1:
-			if !applicationCredentialAvailable(nameMatches[0]) {
-				return client.Credential{}, fmt.Errorf(
-					"GitHub credential %q is not available",
-					nameMatches[0].Name,
-				)
-			}
 			return nameMatches[0], nil
 		default:
 			return client.Credential{}, withExitCode(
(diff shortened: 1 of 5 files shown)


CodeStaple:
Both findings fixed in b9136c1:

- **[low] Coverage fetch failure indistinguishable from success** — a failed read now says so on a human run ("Could not check the GitHub App's repository access (…); continuing.") and the add proceeds, because a failed check is not a verdict about this repository and the platform refuses the clone later if the installation truly cannot read it. A structured run (`-o json`) prints nothing there, so its output stays parseable.
- **[low] Account login interpolated into the URL unescaped** — the organisation login is now `url.PathEscape`d. It is a path segment of a URL printed for a human to click and it arrives from the platform rather than from this process, so it is escaped rather than trusted to be a bare GitHub login.

CodeStaple:
On the re-review of b9136c1. That commit had accidentally swept in the `pipeline run` working-directory defaults, which belong to #264; f6cf20c restores this branch to the credential change alone (`cmd/application.go` and its test), so the `cmd/pipeline*.go` findings are answered there rather than here.

- **[medium] `credentialReachesRepository` may dereference a nil apiClient** — FIXED in f6cf20c: it answers "no opinion" when no client is wired, the way the pipeline lane's inference already does.
- **[medium] New client method and helper not visible in the diff** — `GetCredentialRepositories` is existing client code (`internal/client/credentials.go:132`, added with `ankra credentials repositories`); this branch only calls it, so the diff is scoped to `cmd/`.
- **[medium] `sort` and `net/url` imports not shown** — both are in the file: `net/url` was already imported and is used by the remote parser, and `sort` is imported for `sortCredentialsByAvailability`. `go build`, `go vet` and `golangci-lint` are green on the head, which is what would catch a missing import.
- **[low] Inferred HEAD sha used when `--application` names a different repository** (now #264's code) — the platform resolves a dispatched sha against the repository's refs before the run is created (`verifyRunProvenance`, `usecase/pipelines/runs.go:226`), so a commit that does not belong to the named application's repository is refused there rather than run. The inference is also what makes the common case work: a repository with two applications must pass `--application`, and the commit still comes from the checkout in front of the user.

PR #264, opened 2026-09-07
feat(pipeline): run and list read the application and the commit from the checkout (ankra-ctsmd)

CodeStaple:
## What this fixes

Dispatching a run meant assembling two facts the shell already had:

```
ankra pipeline run --application shipfortune --sha $(git rev-parse HEAD)
```

Both now default from the working directory:

```
$ ankra pipeline list
Using the application bound to CodeStaple/vela (pass --application to choose another).
╭──────────────────────────────────────┬───────┬───────────┬─────────┬─────────────────┬──────────────╮
│ ID                                   │ RUN # │ STATUS    │ TRIGGER │ REF             │ SHA          │
```

- **The application** is the one bound to the checkout's `origin` repository, matched on owner and name against the organisation's applications.
- **The commit** is the checkout's `HEAD`, and the ref defaults to the checked-out branch (a detached HEAD contributes a sha and no ref, which is exactly what a dispatch wants).

## What it does not do

Nothing is inferred silently: each inference prints what it used, on **stderr**, so a structured `-o json` output stays exactly what a caller can parse.

Nothing is guessed either. Outside a checkout, with no `origin` remote, or when no application is bound to the repository, the command asks for the flag exactly as before. When the repository has **several** applications the refusal names them rather than repeating the generic line:

```
Error: CodeStaple/shipfortune has 2 applications in this organisation (shipfortune, shipfortune2); pass --application to say which
```

The dispatch contract is unchanged. A run is still dispatched at one named commit, because resolving a ref to a commit belongs to the trigger lane and a dispatch must never run against whatever commit the platform stored last. What changed is only who reads the sha: the CLI, from the checkout in front of it, instead of the user pasting it back.

## Tests

`TestPipelineRunReadsTheWorkingDirectoryHead` (a seeded checkout dispatches its HEAD and branch) and `TestPipelineRunRequiresSHA` re-pointed to run from a directory that is not a repository, which is where that contract still holds. Full `go test ./cmd/ ./internal/...` and `golangci-lint` green via the pre-commit hook.

Verified live against the Ankra - Development organisation: the two outputs above are real, from `CodeStaple/vela` (one application) and `CodeStaple/shipfortune` (two).

Bead: ankra-ctsmd

### DIFF
--- a/cmd/pipeline.go
+++ b/cmd/pipeline.go
@@ -151,9 +151,15 @@ func pipelineSelectorFromWorkingDirectory(requestContext context.Context) (clien
 	fullName := repository.Owner + "/" + repository.Name
 	matchedIDs := []string{}
 	matchedNames := []string{}
+	listingExhausted := false
+	// The listing is walked unfiltered: the server-side `search` matches an
+	// application's NAME, and an application is free to be called something
+	// other than the repository it builds, so filtering by the repository
+	// name would silently miss exactly those. The walk is bounded the same
+	// way resolveApplicationID's is.
 	for page := 1; page <= maxApplicationLookupPages; page++ {
 		payload, listError := apiClient.ListApplicationsRaw(
-			requestContext, page, maxApplicationLookupPageSize, repository.Name)
+			requestContext, page, maxApplicationLookupPageSize, "")
 		if listError != nil {
 			return client.PipelineSelector{}, "", nil
 		}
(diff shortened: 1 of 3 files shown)


CodeStaple:
All three fixed in 02278b1:

- **[high] `--ref` without `--sha` paired the user's ref with the local HEAD** — real bug, and thank you. A ref the user named is now the whole question: the checkout is read only when no `--ref` was given, so running `release-2.0` from a checkout sitting on `main` asks for `--sha` exactly as before instead of dispatching main's commit under the release ref. Pinned by `TestPipelineRunDoesNotPairANamedRefWithTheLocalHead`.
- **[low] Search filtered on application name, not repository name** — correct, and it would have silently missed every application named differently from the repository it builds. The listing is now walked unfiltered, bounded by the same page cap `resolveApplicationID` uses, matching on `app_repo_owner` / `app_repo_name`.
- **[low] Lookup failures indistinguishable from "no application"** — a walk that runs past the page cap now answers nothing, so a partly-read listing is never reported as an absence; the caller falls through to the usual "one of --application or --repository is required". A listing error stays a silent no-answer on purpose: the fallback message is already the actionable one, and the flag the user is about to pass makes the lookup moot.

CodeStaple:
On the re-review of 02278b1:

- **[low] HEAD inference depends on helpers not shown in the diff** — `localHeadCommit` is in this diff (`cmd/pipeline_run.go`); the one helper it leans on, `executeGit`, is existing CLI code in `cmd/application.go`, which is why it does not appear here. `go build`, `go vet`, the package tests and `golangci-lint` are green on this head.

PR #265, opened 2026-09-07
feat(security): history and report-schedule commands for a cluster ankra-1lzs5.4

CodeStaple:
## Why

The Security Center parity review (epic ankra-1lzs5) left one family on the browser only: the per-cluster Trivy integration's history trend and the scheduled email report. cluster#2789 gives those routes bearer twins; this PR gives them commands.

## What

- `ankra security history --cluster <c> [--days N]` (default 90; the platform clamps 1..365): the daily snapshots as a table (findings by severity, actionable, fixable severe, workloads, namespaces, risk) oldest first, plus a one-line trend between the first and last snapshot. A day without a snapshot is absent, never zero; an empty window says the trend is unknown.
- `ankra security report-schedule --cluster <c>`: frequency, recipients, enabled or paused, last sent, next due, or "not configured".
- `ankra security report-schedule set --cluster <c> --frequency weekly|monthly --recipient <email>... [--disabled] [--yes]`: creates or replaces the schedule; recipients are trimmed and de-duplicated, at most 20, at least one while enabled; bad frequency or a non-address is a usage error before the prompt.
- `ankra security report-schedule send --cluster <c> [--yes]`: queues one report now.
- All four take `-o json|yaml`; the embedded `ankra-security` skill and the changelog list them.

## Route census

`TestClusterRoutesAreRegistered` checks every `/api/v1` literal against cluster main's census. The three new paths are not registered on main until cluster#2789 merges, so they are recorded on the allowlist with a header note. Once #2789 lands they resolve and the next `ANKRA_CLUSTER_ROUTES_UPDATE=1` run drops them (verified locally against that branch's census).

Gates: gofmt, `go vet ./...`, `go build ./...`, `go test ./...` (6 packages, 14 new tests), `make lint` 0 issues.

Bead: ankra-1lzs5.4

https://claude.ai/code/session_014pmkLqticsK53bPdD5k1xD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,22 @@
 # Ankra CLI Changelog
 
+## Unreleased
+
+### Added
+
+- **`ankra security history --cluster <c> [--days N]` shows how a cluster's
+  posture moved.** The daily snapshots the portal charts - findings by
+  severity, what is actionable once dispositions apply, workloads and
+  namespaces covered, the risk score - as a table oldest first, with the
+  trend between the first and last snapshot. A day without a snapshot is
+  absent, never zero; an empty window says the trend is unknown.
+- **`ankra security report-schedule` manages the emailed security report.**
+  The bare command shows a cluster's schedule (frequency, recipients,
+  enabled, last sent, next due) or "not configured"; `set --frequency
+  weekly|monthly --recipient <email>...` creates or replaces it (`--disabled`
+  stores it paused); `send` queues one report now. Both writes confirm unless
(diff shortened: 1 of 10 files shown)


CodeStaple:
Addressing the review (follow-up commit pushed):

- **Snapshot order.** The window is now copied and sorted by date before rendering, so the table and the trend read oldest to newest whatever order the wire uses; the rendering test feeds the points newest first.
- **Single snapshot.** Detected by `len(points) == 1`, not by equal dates.
- Also dropped the three route-census allowlist entries: cluster#2789 merged, so the paths resolve against main's census now.

PR #266, opened 2026-09-07
feat(credentials): the detail view says whether a credential is usable, and why not (ankra-ctsmd)

CodeStaple:
## What this fixes

`ankra credentials list` prints a state column, so a credential can read `down` there. `ankra credentials get` then showed nothing about it at all:

```
Credential Details:
  ID:       821a0ed4-4c96-41c2-9190-fc6ca5f2f845
  Name:     github-app-codestaple
  Provider: github
  Created:  1 week ago
```

No state, no availability, no reason, and no hint that a third command holds the answer. The reason lives on the repository-coverage read, which a user has to already know exists.

## After

```
Credential Details:
  ID:       821a0ed4-4c96-41c2-9190-fc6ca5f2f845
  Name:     github-app-codestaple
  Provider: github
  Created:  1 week ago
  Usable:   no
  Reason:   The GitHub App installation behind this credential cannot reach taelio/status-page-next-go, so every build and deploy for the applications bound to it will fail with a 404. Grant the installation access to those repositories on GitHub, or point the applications at the credential whose installation already has it.
            Run 'ankra credentials repositories github-app-codestaple' for the repository list.
```

That output is real, from the Ankra - Development organisation.

## Notes

- `CredentialDetail` gains the `Available` field the route has always sent and nothing read.
- The coverage read happens only for a GitHub credential that is already unusable, so a healthy credential costs no extra request and prints one line.
- A coverage read that fails leaves the verdict without a reason rather than failing the command: the user still learns the credential is unusable.

## Tests

`TestCredentialsGetExplainsAnUnusableCredential` and `TestCredentialsGetSaysAUsableCredentialIsUsable`, both against an httptest server answering the detail and coverage routes. Full `go test ./cmd/ ./internal/...` and `golangci-lint` green via the pre-commit hook.

Bead: ankra-ctsmd

### DIFF
--- a/cmd/credentials.go
+++ b/cmd/credentials.go
@@ -289,11 +289,22 @@ func printCredentialHealth(credential *client.CredentialDetail) {
 		return
 	}
 	fmt.Printf("  Usable:   no\n")
+	// Only a GitHub App credential has a recorded reason to show: repository
+	// coverage is an App-installation concept, and no other provider answers
+	// the route this reads. A credential of another provider stops at the
+	// verdict rather than inventing an explanation for it.
 	if apiClient == nil || !strings.EqualFold(credential.Provider, "github") {
 		return
 	}
 	coverage, coverageError := apiClient.GetCredentialRepositories(credential.ID)
-	if coverageError != nil || coverage == nil || strings.TrimSpace(coverage.CoverageMessage) == "" {
+	if coverageError != nil {
+		// The reason could not be read. Saying so beats printing "no" and
+		// stopping, which reads as "there is no reason recorded".
+		fmt.Printf("  Reason:   could not be read (%v)\n", coverageError)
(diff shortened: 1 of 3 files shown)


CodeStaple:
Both findings addressed in 5cf3dd5:

- **[low] Coverage fetch failure indistinguishable from "no reason recorded"** — FIXED. A failed read now prints `Reason: could not be read (…)` with the command that asks again. The verdict itself still stands, because whether Ankra can use the credential comes from the credential, not from the coverage read. Pinned by `TestCredentialsGetSaysWhenTheReasonCannotBeRead`.
- **[low] Reason lookup restricted to the GitHub provider** — deliberate, and now stated where it lives. Repository coverage is a GitHub App installation concept: `GET /org/credentials/{id}/repositories` answers for an App-backed credential, and no other provider records a reason there. A credential of another provider stops at the verdict rather than inventing an explanation for it, which is still strictly more than the view said before this PR, where it printed no health at all.

PR #271, opened 2026-09-08
fix: show application members in ankra cluster stacks list

CodeStaple:
## What

`internal/client.ClusterStackListItem` decodes only `manifests` and `addons`. The platform's stack listing has always returned a third member array, `applications`, beside them.

A stack deployed from an Ankra application is frequently nothing *but* application members. For such a stack the CLI printed:

```
Manifests:   0
Addons:      0
```

…listed no members in the detail view, and dropped the applications entirely from `-o json`. That output reads as an empty stack. It is what sent a session looking for a stack that was never broken, and then concluding that cloning it would produce an empty draft (the clone lane already reports `applications_cloned` correctly — cluster#1971).

## Change

- `ClusterStackListItem` gains `Applications []StackApplication`, and `StackApplication` carries name, namespace, `platform_application_id` / `platform_application_version`, parents, state and health. The nullable wire members decode to the empty string rather than failing the listing.
- `ankra cluster stacks list` gains an **Applications** column.
- `ankra cluster stacks list <name>` counts applications alongside manifests and addons, and lists them with namespace, state, parent edges and the `<application-id>:<version>` they deploy. The detail labels are re-padded one column so the new, longer label still aligns.
- The per-member state glyph moves into `stackMemberStateIcon` instead of being written a third time.

## Tests

- `TestListClusterStacksDecodesApplications` — a raw platform response body with an application-only stack decodes with its member intact.
- `TestListClusterStacksToleratesNullApplicationMembers` — `null` namespace/version/health does not fail the listing.
- `TestClusterStacksListShowsApplicationMembers` — the detail view of an application-only stack shows the count, the name and the application reference.

## Validation

- `go build ./...`
- `go test -count=1 ./...` — all packages ok
- `golangci-lint run` — 0 issues
- Pre-commit gate green (tests + lint).

## Related

`ankraio/cluster#2823` fixes the matching backend gap: the draft-source lane of the cross-cluster stack clone dropped application members outright.

## Bead

Not filed: the shared Dolt tracker is in an unresolved merge conflict (an id collision on `ankra-vn0bd.3.14` needing the SETUP.md operator merge), so `bd create` fails on this machine.

Bead: ankra-l84qb


Myothas:
## Shepherd review: ready once the CHANGELOG is re-anchored — and it has no backend dependency

The change is correct, additive and well tested. The only thing standing between it and merge is a stale CHANGELOG anchor.

**Conflict is CHANGELOG.md only.** `merge-tree` shows `cmd/cluster_stacks.go`, `cmd/e2e_test.go`, `internal/client/stacks.go` and `stacks_test.go` all auto-merging. The hunk was anchored on an `## Unreleased` block that master has since folded under v0.15.1 / v0.16.0-rc0-rc2. Re-anchor under a fresh `## Unreleased` and it is clean. It also does not collide with #275 — that edits `cluster_stacks.go` at 422-485 (the resolver), this one at 64-195 (the renderer).

**You asked whether this waits on cluster#2823 — it does not.** The `applications` field already exists in the platform response: cluster's own parity test decodes it beside `manifests`/`addons` from the live `/stacks` listing (`go/internal/importedapi/postgres_parity_render_test.go:731`), and there is a rendered body with `"applications":[]` at `go/internal/agentictools/postgres_parity_test.go:245`. #2823's own body says "The API already returns both." #2823 fixes a different lane — the draft-source branch of the cross-cluster clone — and has now merged (`0edaee27`) independently. Merging this shows real data immediately; it neither errors nor renders empty.

Also worth knowing: bead ankra-l84qb's other half (`CloneStackToClusterResult` dropping `applications_cloned`) is **already on master** (`internal/client/stacks.go:227`, landed in #175), so merging this genuinely completes the bead.

### Nits
- The detail-view labels were re-padded by one column ("Name:        " → "Name:         ") for the longer "Applications:" label. Anyone doing fixed-column `cut` on that output breaks. Harmless for `-o json`/`-o yaml`, which go through `renderStructured` before the text path.
- `Applications` has no `omitempty`, so `-o json` now always emits `"applications": null` against a platform that omits the key. Consistent with the existing `Manifests`/`Addons`, and `jq '.applications | length'` still yields 0 — cosmetic.
- Not tested: the new **table** column (only the detail view is asserted), and `formatApplicationReference`'s `"-"` branch when `platform_application_id` is empty.

The three new tests are good and behavioural — `TestListClusterStacksToleratesNullApplicationMembers` is the valuable one, since the backend really does render member `state` as `*string` (`postgres_parity_render_test.go:731-734`), so a `null` that failed the whole listing would be a live defect.

Rebase the CHANGELOG and this is good to go.

PR #272, opened 2026-09-08
fix(cluster encrypt): seal the whole encrypted_paths declaration, and stop calling every 422 a git push failure (ankra-bfvfy, PLA-830)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-830/1152-cluster-encrypt-manifest-cluster-mode-set-fails-update-stack-500
Bead: ankra-bfvfy

## What

Three fixes to the `ankra cluster encrypt` path, all one failure: a secret change the customer could not make, refused with a message that pointed at the wrong thing.

1. **Cluster mode seals the whole declaration.** `encrypt manifest` / `encrypt addon` send the platform's encrypt route the union of the `--key`s and every `encrypted_paths` entry the resource already declares (requested keys first, deduplicated by key name), and say so on stdout when the declaration added anything. Only the requested keys are verified afterwards, as before.
2. **File mode does the same.** It had the identical defect one lane over.
3. **A 422 is no longer blanket-labelled "git push failed".**

## Why

PLA-830 (Depict): `ankra cluster encrypt manifest runway-envs --cluster prod-multi-az --key EMBEDDING_VOYAGE_MERCHANT_IDS --set '...'` printed the in-memory `--set` and encrypt lines, then

```
Error: update stack failed: status 500, body: Internal Server Error
```

The escaped commas were fine and the stack was fine. Confirmed from the prod log for both runs (`ankra-cluster-api-go`, 10:02:39Z and 10:03:16Z, `PATCH /api/v1/org/clusters/imported/e781.../stacks/runway` 500):

```
Cannot persist values for 'runway-envs': the document carries SOPS metadata but declared
encrypted paths [stringData.METRICS_AUTH_TOKEN stringData.RUNWAY_TYPESENSE_ADMIN_KEY
 stringData.SEARCH_TOKEN_SIGNING_KEY stringData.TINYBIRD_EVENT_WRITING_KEY
 stringData.TRINITY_DB_PASSWORD stringData.TYPESENSE_SHOPIFY_AUTH_HEADER
 stringData.TYPESENSE_SHOPIFY_HOST stringData.TYPESENSE_SHOPIFY_PORT
 stringData.TYPESENSE_SHOPIFY_PROTOCOL stringData.TYPESENSE_SHOPIFY_WRITE_HOSTS
 stringData.VOYAGE_API_KEY] are plaintext. Re-encrypt the document with sops (fresh MAC) before saving.
```

The stored manifest content was plaintext under paths declared earlier (a portal or API save leaves a declared value plaintext until the next GitOps push seals it), and the CLI sealed only `EMBEDDING_VOYAGE_MERCHANT_IDS`. The result is a SOPS document with plaintext under the other eleven declared paths. The push lane passes a SOPS document through byte-for-byte rather than sealing it, so the store-time guard (cluster#2647) refuses that save — correctly: that shape is the one that leaked live S3 credentials to GitHub on 2026-07-07.

So the guard is right and the CLI was wrong. **`encrypt manifest --key X` in cluster mode could never succeed on a manifest that declared more than one encrypted path** — the declaration, not the document, is what the platform judges a save against, and the CLI only ever sealed one side of it. The platform's encrypt route unions in extra paths *only* when the input already carries sops metadata (`EncryptContent`), which a plaintext stored document does not.

**File mode (2)** reaches the same shape by a different route: `ankra cluster decrypt manifest -f` writes the plaintext document back to disk and deliberately leaves `encrypted_paths` declared, so the documented decrypt-edit-encrypt loop leaves a plaintext file under a multi-entry declaration. Sealing only `--key` there is refused on the next `ankra apply`.

**The 422 label (3)** is the "could the person act on it" half. cluster#2828 (same bead, merged 2026-09-08) stopped the guard's verdict reaching the CLI as a bare 500 and made it a 422 carrying the sentence — but `mapPatchError` prefixes every 422 with `git push failed:`. Seven different refusals answer 422 on the stack-write routes and only one is a push failure; the SOPS store guards, spec validation, the secret-slot pre-flight, a circular dependency and a rejected name are all refused *before* the database write and before git is touched. The customer would have been told their git push failed, in the same breath as confirming no commit was made — sending them to look at GitOps credentials for a refusal that never reached git. The platform already marks the real one with `error_code: GIT_PUSH_FAILED`, so that now decides; every other refusal is surfaced as the sentence the platform wrote.

A sealed stored document is unaffected: the encrypt route already merges the paths it observes as ciphertext, and the union is a superset.

## Tests

`cmd/cluster_encrypt_declared_paths_test.go`:
- manifest with declared `[stringData.OTHER_SECRET, data.username]` and `--key password --set ...` sends `[password stringData.OTHER_SECRET data.username]` and announces it; the PATCH keeps the declared entries and adds `password`
- a `--key` naming an already-declared key is not duplicated (`data.username` and `username` are one key)
- no declaration: only the requested keys are sent, nothing announced (existing behaviour, pinned)
- addon with declared `[smtpPassword]` and `--key adminPassword` sends `[adminPassword smtpPassword]`
- **file mode**: manifest and addon with declared paths seal the union; no declaration still seals only the keys
- **the store-guard refusal reaches the user as itself** — asserts the error carries no "git push" and equals the platform's verdict verbatim
- `mapPatchError`: a `GIT_PUSH_FAILED`-marked 422 keeps the git prefix, an unmarked one does not

`cmd/cluster_partial_test.go` / `cmd/upgrade_e2e_test.go`: the two cases that pinned the old blanket prefix now send the `error_code` the platform sends, plus a new case for the SOPS verdict.

Gates: the repo pre-commit hook ran the full suite and golangci-lint on the final commit — all packages ok, **0 lint issues**.

Not on the scary list: CLI behaviour in one command family. Ships with the next release tag.

https://claude.ai/code/session_017jU4skEXGw73M7eRSMYBPK

### DIFF
--- a/cmd/cluster_encrypt.go
+++ b/cmd/cluster_encrypt.go
@@ -170,7 +170,10 @@ Two modes:
 
   File mode (-f cluster.yaml): rewrite a local cluster.yaml's referenced
     from_file in place, adding the key to encrypted_paths in the file. Used by
-    GitOps workflows where the source of truth is on disk.
+    GitOps workflows where the source of truth is on disk. Every key the file
+    already declares in encrypted_paths is sealed together with --key, so a
+    document decrypted with "ankra cluster decrypt" (which leaves the
+    declaration in place) is sealed whole again.
 
 In cluster mode, --set applies value edits in-memory BEFORE encrypting, so the
 new secret value and its encryption land in a single commit - the plaintext
@@ -243,7 +246,8 @@ Two modes:
     partial-stack PATCH endpoint. The owning stack is resolved automatically.
 
   File mode (-f cluster.yaml): rewrite the local addon values file referenced
-    by the cluster.yaml in place, adding the key to encrypted_paths.
+    by the cluster.yaml in place, adding the key to encrypted_paths. Keys the
+    file already declares are sealed together with --key.
 
 Examples:
   # Cluster mode against the selected cluster
@@ -669,7 +673,8 @@ func runEncryptManifestFile(cmd *cobra.Command, manifestName string, leafKeys []
 
 	fmt.Printf("Encrypting %s in manifest %q...\n", describeEncryptKeys(leafKeys), manifestName)
 
-	encryptedContent, err := apiClient.EncryptYAML(string(manifestContent), leafKeys)
+	pathsToSeal := pathsToSealWithDeclared(encryptAnnounceWriter(cmd), leafKeys, foundManifest.EncryptedPaths)
+	encryptedContent, err := apiClient.EncryptYAML(string(manifestContent), pathsToSeal)
 	if err != nil {
 		return fmt.Errorf("encryption failed: %w", err)
 	}
@@ -787,7 +792,9 @@ func runEncryptAddonFile(cmd *cobra.Command, leafKeys []string) error {
 
 	fmt.Printf("Encrypting %s in addon %q...\n", describeEncryptKeys(leafKeys), encryptAddonName)
 
-	encryptedContent, err := apiClient.EncryptYAML(string(addonContent), leafKeys)
+	pathsToSeal := pathsToSealWithDeclared(encryptAnnounceWriter(cmd), leafKeys,
+		getEncryptedPathsFromConfig(foundAddon.Configuration))
+	encryptedContent, err := apiClient.EncryptYAML(string(addonContent), pathsToSeal)
 	if err != nil {
 		return fmt.Errorf("encryption failed: %w", err)
 	}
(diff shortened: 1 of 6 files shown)


Myothas:
@ankraai Thanks — two taken in 23ef31b, one answered.

**[high] panic on repeated `--key`** — taken as a code change, but the diagnosis does not hold: a panic was never reachable. The announcement is guarded by `len(merged) > len(leafKeys)` and the panic needs `len(leafKeys) > len(merged)`; those cannot both be true. And `normalizeAndAnnounceEncryptKeys` already deduplicates by leaf key before the helper is reached, so `leafKeys` is duplicate-free at both call sites today.

The underlying point is right, though, and it is a real (if latent) bug rather than a crash: with repeated keys the offset shifts and the line names the *wrong* declared entries. `pathsToSealWithDeclared` is unit-tested directly and shouldn't depend on an invariant maintained three frames away, so the added entries are now computed by key name instead of by slice arithmetic, with a regression test (`--key password --key data.password` are one key).

**[low] announcement overstates a stale entry** — agreed and fixed. It now reads "Including the already-declared `X` in the same encrypt, ..." rather than "Also sealing", which is what actually happens: the entry is passed to the encrypt route and may select nothing.

**[medium] unmarked genuine push failures** — checked on the platform side; not taking a change. Every git-push refusal on these routes goes through `writeGitPushFailure`, which is `httpjson.WriteDetailWithErrorCode(..., failure.ErrorCode())`, and `GitPushFailure.ErrorCode()` returns `GIT_PUSH_DEFERRED` for a designed refusal and `GIT_PUSH_FAILED` for every other case — there is no path that emits a push failure without a code. Both twins (`cliapi`, `importedapi`) use that one writer. The contract is already stated in `internal/client/gitpush.go`: "Both codes are always emitted together."

Two things make the trade clearly favourable even in the body-rewriting case you raise. The old blanket prefix was correct for **one** of the seven refusals that answer 422 on these routes and wrong for the other six — including the SOPS store guard, which is the one this PR exists for; a customer who had just confirmed no commit was made was going to be told their git push failed. And a genuine push failure's detail already names git itself ("...but pushing the change to the Git repository failed. The Git repository is out of sync."), so an unmarked one still reaches the user as a git problem — it loses a redundant prefix, not the diagnosis.

PR #274, opened 2026-09-09
fix(apply): read and send manifests[].force and auto_remediate from the file (ankra-cbktk, PLA-834)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-834/1156-cli-cluster-apply-stores-forcefalse-on-manifests-regardless-of
Bead: ankra-cbktk

## Problem

`ankra cluster apply -f` dropped a manifest's `force` (and `auto_remediate`) flag. At v0.15.0 `client.Manifest` (`internal/client/clusters.go:304`) has no such field and `buildManifest` (`cmd/apply.go:732`) never reads the keys, so the `manifests[]` entry on the wire says nothing about force and the platform stores false.

On Smartoptics `so-development` (PLA-834, 2026-09-09) a GitOps PR set `force: true` on ten strimzi CRD manifests and the sync succeeded; five minutes later CI re-applied the unchanged stack file (which also carries `force: true`) with the CLI and the platform exported `force: false` back into the cluster file. Reproduced by the customer on v0.8.0 and v0.14.0; the code is the same at v0.15.0.

## Fix

- `client.Manifest` gains `Force` and `AutoRemediate`, always sent (no `omitempty`): apply is declarative, so a file without the key means false, the same rule the GitOps cluster file follows.
- `buildManifest` reads both through `parseBoolField`: absent or null is false, a non-boolean (`force: "true"`) is an error naming the key, like `parseGroupField`, rather than a silent off.

Left alone on purpose:

- `ManifestSpec` (partial patches behind `manifests update`, `manifests create`, `cluster encrypt`): those lanes patch one field of a manifest read back from the API, and the read shape has no force, so an always-sent field there would clear a stored flag. They rely on the platform's absent-means-keep rule from ankraio/cluster#2843 (bead ankra-tg8zb).
- `cluster clone`: it writes the parsed YAML node back, so keys its structs do not model already survive a copy.

## Behaviour note for the reviewer

With this CLI the file is authoritative in both directions: a file that carries `force: true` sets it, and a file that lacks the key clears it, exactly as the GitOps lane treats the cluster file. Before this PR the CLI could only ever clear. Older CLIs (through v0.15.0) keep clearing until cluster#2843 deploys, after which they leave the stored flag alone.

## Tests

`cmd/apply_flags_test.go`: both flags pass through and marshal as `"force":true`; absent flags are false and still on the wire; null reads as false; a quoted `force` and a numeric `auto_remediate` are rejected with the key named. The new test does not compile against the previous struct, which is the "fails before" for a dropped field.

## Gates run locally

`gofmt -l`, `go build -buildvcs=false ./...`, `go test -buildvcs=false -count=1 ./cmd/ ./internal/client/`, `go vet` on both packages. `golangci-lint` is not installed here; the full linter runs in CI.

## Merge notes

Not on the scary list: three files, under 100 lines, no auth, protocol, CI or deploy surface. Ships with the next CLI release; the customer's CI pins v0.8.0, so cluster#2843 is what protects them until they upgrade.

### DIFF
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -742,7 +742,10 @@ func parseBoolField(m map[string]interface{}, key string) (bool, error) {
 	}
 	value, ok := raw.(bool)
 	if !ok {
-		return false, fmt.Errorf("'%s' must be true or false (got %v)", key, raw)
+		// Name the type as well as the value: force: "true" and force: true
+		// both print as true, and the quoted string is the case this guard
+		// exists for.
+		return false, fmt.Errorf("'%s' must be true or false (got %v of type %T)", key, raw, raw)
 	}
 	return value, nil
 }
--- a/cmd/apply.go
+++ b/cmd/apply.go
@@ -729,6 +729,24 @@ func parseGroupField(m map[string]interface{}) (string, error) {
 	return group, nil
 }
 
+// parseBoolField extracts an optional boolean key such as a manifest's
+// 'force' or 'auto_remediate'. Absent (or explicitly null) is false, the
+// declarative reading the GitOps cluster file gets too. A non-boolean value
+// is an error rather than false: the platform types both as booleans and
+// would answer 422, and quietly reading "true" (a quoted string) as off is
+// the same silent drop that lost the flag in the first place (ankra-cbktk).
+func parseBoolField(m map[string]interface{}, key string) (bool, error) {
+	raw, present := m[key]
+	if !present || raw == nil {
+		return false, nil
+	}
+	value, ok := raw.(bool)
+	if !ok {
+		return false, fmt.Errorf("'%s' must be true or false (got %v)", key, raw)
+	}
+	return value, nil
+}
+
 func buildManifest(mm map[string]interface{}, baseDir string) (client.Manifest, error) {
 	name, _ := mm["name"].(string)
 	if name == "" {
@@ -801,12 +819,26 @@ func buildManifest(mm map[string]interface{}, baseDir string) (client.Manifest, 
 		return client.Manifest{}, err
 	}
 
+	// The apply flags. Dropped on the floor until ankra-cbktk: a file that
+	// said force: true applied without it, and the platform stored false over
+	// the flag the customer had set through a GitOps PR (PLA-834).
+	force, err := parseBoolField(mm, "force")
+	if err != nil {
+		return client.Manifest{}, err
+	}
+	autoRemediate, err := parseBoolField(mm, "auto_remediate")
+	if err != nil {
+		return client.Manifest{}, err
+	}
+
 	return client.Manifest{
 		Name:             name,
 		ManifestBase64:   encoded,
 		Namespace:        ns,
 		Parents:          parents,
 		EncryptedPaths:   encryptedPaths,
+		Force:            force,
+		AutoRemediate:    autoRemediate,
 		Group:            group,
 		AgentsMd:         agentsMd,
 		AgentsMdFromFile: agentsMdFromFile,
--- a/cmd/apply_flags_test.go
+++ b/cmd/apply_flags_test.go
@@ -1,0 +1,110 @@
+package cmd
+
+import (
+	"encoding/json"
+	"strings"
+	"testing"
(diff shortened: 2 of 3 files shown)


Myothas:
**🤖 tech-agent** — Ankra AI review (commit cd496fe8618a): both findings answered

- **[medium] Force/AutoRemediate now sent as false by every Manifest constructor** — no change needed; `buildManifest` is the only producer, and every lane the struct travels starts from the file. Census of the tree at 035b119e0e39: the only `client.Manifest{...}` literal outside tests is the return in `buildManifest` (`cmd/apply.go:834`); the only `[]client.Manifest` is the slice `buildStack` fills from it (`cmd/apply.go:542`); and `buildStack` is called once, from the file loader behind `buildImportRequest` (`cmd/apply.go:267`). The three request paths that marshal it, `ApplyCluster` (`cmd/apply.go:185`), `CreateStackDraft` (`cmd/cluster_draft.go:77`) and `ValidateCluster` (`cmd/cluster_validate.go:66`), each take the spec `buildImportRequest` built from `-f`, and `cmd/cluster_draft.go:151` sends an empty `Stacks` list. The `make([]Manifest, 0)` sites in `internal/client/clusters.go:446,504,572` only turn a nil slice into an empty one. `client.Manifest` is never a decode target: none of the `json.Unmarshal` sites in `internal/client/clusters.go` fill a `Stack`- or `Manifest`-bearing type, so nothing reads a manifest back from the API and re-sends it through this struct. The partial-patch lanes (`manifests update`/`create`, `cluster encrypt`) use `ManifestSpec` in `internal/client/iac.go`, which this PR deliberately leaves without the field for exactly the reason the finding names, and `cluster clone` writes the parsed YAML node back through `ManifestConfig`. So "clearing is intended everywhere" holds trivially: everywhere is the file-driven apply/draft/validate lane, where the file is authoritative, and the UI-set flag the finding worries about is protected in the patch lanes by the platform's absent-means-keep rule (ankraio/cluster#2843).
- **[low] Error message loses type information for the offending value** — fixed in 035b119e0e39. The rejection now reads `'force' must be true or false (got true of type string)`; both rejection subtests in `cmd/apply_flags_test.go` assert on the type (`of type string`, `of type int`). `gofmt -l`, `go build -buildvcs=false ./...`, `go test -count=1 ./cmd/ ./internal/client/` and `go vet` on both packages are clean locally.

Myothas:
**🤖 tech-agent** — Ankra AI review (commit 035b119e0e39): one finding, answered

- **[low] Other Manifest construction sites now always send force/auto_remediate as false** — no change needed; this is the same question as the medium finding on cd496fe8618a, and the construction-site census in the previous comment answers it for this head too: `buildManifest` (`cmd/apply.go:834`) is the only producer of `client.Manifest`, it is reached only through `buildStack` from the `-f` file loader, and the struct is never decoded from an API response. There is no import, convert or addon-to-manifest path that builds a `client.Manifest`: `grep -rn 'client\.Manifest{\|\[\]client\.Manifest' cmd internal` outside tests returns only `cmd/apply.go`. The three commands that send it (`cluster apply`, `cluster draft`, `cluster validate`) all take the spec `buildImportRequest` built from the file, so "the file is authoritative" is the rule everywhere the field is on the wire; UI-set flags in the partial-patch lanes go through `ManifestSpec`, which this PR leaves alone.

Myothas:
Stacks lane note (round r-20260909T070153Z-1892db): the 'absent flags are false and still sent' contract here and cluster#2843's 'omitted keeps the stored flag, present false clears' rule were built against each other. Once this ships, a file without force: sends "force":false, which the platform now treats as an authoritative clear, so a stored force: true set through Git is cleared by a routine apply: the PLA-834 outcome again. Tracked as ankra-mp2tr with a ruling card to Mattias (d-1788937936937423000); recommendation is *bool + omitempty so an omitted key is omitted on the wire and an explicit false still clears. Unreleased today (master = v0.15.0 + this commit), so please hold the next release until the ruling lands. This PR cites ankra-cbktk, which no reachable tracker holds; ankra-mp2tr is the record.

PR #275, opened 2026-09-09
fix(cluster): every cluster-scoped command takes the cluster's name, not just its id (ankra-aprvp)

Myothas:
## Why

`ankra cluster list` is the one place a user reads a cluster's identity, and it
shows the **name**. Until now only three commands accepted one.

ankra-cli#239 taught the `cluster playground` verbs to resolve a name to the id
their routes want. They were the only group. A sweep of `master` found **103
more commands** whose positional argument is the cluster and which forwarded it
untouched:

| Group | Commands |
|---|---|
| `cluster` | bastion, managed, mesh, node-group, nodes, scale, ssh-keys, upgrade |
| `control-plane` | get, set-count, set-instance-type |
| provider lanes | DigitalOcean, Hetzner, OVH, UpCloud, Scaleway, Proxmox VE, HPE Morpheus |

A name in any of them reached the route as a non-UUID path segment and came
back as a bare `404`, or as a `uuid_parsing` `422` from a query parameter, with
nothing in either to say an id was expected.

## What changed

`resolveClusterArg` is the shared entry point, generalised from the playground
helper it replaces:

- a **UUID short-circuits**, so a scripted caller that already holds the id pays
  no extra request;
- a **name** is resolved through the existing paged `resolveClusterID`;
- an **unknown name** is refused *before* the request, quoting what was typed
  and naming `ankra cluster list`.

Every `Use:` string widened to `<cluster_id|name>`, so the help advertises it.

Two things needed care beyond the mechanical change:

- **Destructive prompts.** A prompt used to name whatever the user typed. Now
  that a name resolves first, the eleven prompts that name the cluster render it
  through `clusterTarget`, which leads with what was typed and appends the id it
  resolved to, rather than confronting the user with a bare UUID.
- **`cluster mesh readiness`** takes several clusters at once. Every argument
  resolves, and each row is still labelled with the name that was typed.

## Tests

New coverage in `cmd/cluster_name_args_test.go`: name resolution (including
case-insensitivity), the id short-circuit costing no listing, the refusal for an
unknown name, the prompt target, and one command from each newly covered shape
(a provider lane, mesh make-ready, mesh readiness).

`TestEveryClusterArgumentAdvertisesTheName` walks the command tree and fails on
any `<cluster_id>` left in a `Use:` string, naming the offending command path,
so a command added later cannot quietly regress to id-only. It was verified to
fail by reverting one usage string.

## Test-fixture churn

The command tests carried placeholder cluster ids (`"cluster-1"`, `"ovh-123"`,
`"c-1"`, `"test-cluster-id"`) that a name-resolving CLI now reads as *names*,
which sent 78 tests to the cluster listing. They use one shared, id-shaped
`testClusterID` instead, which is what the routes actually take. This is the
bulk of the test diff.

## Validation

Local pre-commit gate on the final commit: `go test ./...` green across every
package, `golangci-lint run ./...` reports **0 issues**. `gofmt` is clean across
`cmd/`, including `cluster_mesh_test.go`, which was already unformatted on
`master`.

Bead: ankra-aprvp

https://claude.ai/code/session_01T6Wvx6CJkF6eEwRUhGdxdW

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,34 @@
 # Ankra CLI Changelog
 
+## Unreleased
+
+### Fixed
+
+- **Every cluster-scoped command now takes the cluster's name as well as its
+  id.** The three `ankra cluster playground` verbs learned this in #239, but
+  they were the only group: the other 103 commands - `cluster` bastion,
+  managed, mesh, node-group, nodes, scale, ssh-keys and upgrade,
+  `control-plane`, and every provider lane (DigitalOcean, Hetzner, OVH,
+  UpCloud, Scaleway, Proxmox VE, HPE Morpheus) - forwarded a name verbatim,
+  where it reached the route as a non-UUID path segment and came back as a
+  bare 404 or a `uuid_parsing` 422 with nothing to say an id was expected. A
+  cluster id still passes straight through without the extra lookup, an
+  unknown name is refused before the request, and a destructive prompt names
+  what you typed alongside the id it resolved to.
+- **Name lookup no longer picks one of two clusters whose names differ only by
+  case.** A name typed exactly as the cluster carries it wins immediately;
+  otherwise a single case-insensitive match resolves and several are refused
+  naming both clusters and both ids. This is the resolver the stacks,
+  security, cost and validate commands already shared, so they gain it too.
+- **A cluster name that is 36 characters long with four dashes is looked up
+  rather than forwarded as an id.** The lookup's own shape test was that loose
+  while the caller's was a strict UUID, so names of that shape fell into the
+  gap and answered with the opaque error this release removes everywhere else.
+- **A name beyond the lookup's 5000-cluster paging cap is no longer reported
+  as "not found".** A truncated listing was being presented as a verified
+  absence; the error now says the name was not among the first 5000 and to
+  pass the cluster id instead.
+
 ## v0.15.1 — 2026-09-09
 
 ### Fixed
--- a/cmd/agent_ci_test.go
+++ b/cmd/agent_ci_test.go
@@ -86,7 +86,7 @@ func TestClusterAgentCIGetPrintsTheSettingsAndTheApplyState(t *testing.T) {
 	if runError != nil {
 		t.Fatalf("expected success, got %v", runError)
 	}
-	if mock.getCalls != 1 || mock.lastClusterID != "test-cluster-id" {
+	if mock.getCalls != 1 || mock.lastClusterID != testClusterID {
 		t.Fatalf("expected one read of the selected cluster, got %d for %q", mock.getCalls, mock.lastClusterID)
 	}
 	for _, want := range []string{
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -78,7 +78,14 @@ func resolveCloudProviderNetworking(cmd *cobra.Command) (externalCloudProvider b
 // as `ankra cluster list` shows it.
 //
 // A UUID short-circuits, so a scripted caller that already holds the id pays
-// no extra request. A name is looked up once. Before this existed, only the
+// no extra request. The trade-off is that a cluster whose NAME is itself
+// UUID-shaped cannot be addressed by that name: the argument is forwarded as
+// an id and the route answers 404. Nothing in the CLI or the API constrains a
+// cluster's name, so this is reachable in principle; it stays this way because
+// the alternative costs every scripted caller a listing request on every
+// command, and the cluster id is always an unambiguous way to name it.
+//
+// A name is looked up once. Before this existed, only the
 // three playground verbs resolved a name (ankra-y8l44.35) and every other
 // cluster-scoped command forwarded the name verbatim: it reached the route as
 // a non-UUID path segment and came back as a bare 404, or as a uuid_parsing
--- a/cmd/cluster.go
+++ b/cmd/cluster.go
@@ -72,3 +72,36 @@ func resolveCloudProviderNetworking(cmd *cobra.Command) (externalCloudProvider b
 	}
 	return externalCloudProvider, includeNetworking, nil
 }
+
+// resolveClusterArg turns a positional cluster argument into the id the
+// platform routes want, accepting either the id itself or the cluster's name
+// as `ankra cluster list` shows it.
+//
+// A UUID short-circuits, so a scripted caller that already holds the id pays
+// no extra request. A name is looked up once. Before this existed, only the
+// three playground verbs resolved a name (ankra-y8l44.35) and every other
+// cluster-scoped command forwarded the name verbatim: it reached the route as
+// a non-UUID path segment and came back as a bare 404, or as a uuid_parsing
+// 422 from a query parameter, with nothing in either to say an id was
+// expected (ankra-aprvp).
+func resolveClusterArg(nameOrID string) (string, error) {
+	if isLikelyClusterID(nameOrID) {
+		return nameOrID, nil
+	}
+	clusterID, resolveError := resolveClusterID(nameOrID)
+	if resolveError != nil {
+		return "", fmt.Errorf("%w (pass the cluster's name as `ankra cluster list` shows it, or its id)", resolveError)
+	}
+	return clusterID, nil
+}
+
+// clusterTarget renders the cluster a confirmation prompt is about. The
+// argument the user typed is what they recognise, so it leads; the id it
+// resolved to is appended when they differ, so a destructive prompt names the
+// cluster the API is actually about to be asked to act on.
+func clusterTarget(typed, clusterID string) string {
+	if typed == clusterID {
+		return fmt.Sprintf("%q", typed)
+	}
+	return fmt.Sprintf("%q (cluster %s)", typed, clusterID)
+}
--- a/cmd/cluster_bastion.go
+++ b/cmd/cluster_bastion.go
@@ -313,12 +313,16 @@ func newBastionCmd(opsFn func() bastionOps, provider string, supportsResize, sup
 	}
 
 	statusCmd := &cobra.Command{
-		Use:   "status <cluster_id>",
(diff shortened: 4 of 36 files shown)


Myothas:
Both findings fixed in ebef022.

**Join/leave printed the resolved id.** They now render the cluster through `clusterTarget(args[1], clusterID)`, the same helper the destructive prompts use, so the success line leads with what the user typed and appends the id it resolved to. That makes them consistent with readiness, which already kept the typed label.

**Ratchet regex was too narrow.** It matched two literal spellings. It now pulls every bracketed placeholder out of each `Use` string and flags any token naming a cluster id that lacks the widened `|name` form, so `[cluster_id]`, `<cluster-id>` and `<clusterID>` are all caught. It also reports the offending token next to the command path. Verified by temporarily rewriting `cluster scale` to each of the three spellings in turn: the test fails on all three, and passes once restored.

On the open question about `resolveClusterID`, which the diff does not show: it is pre-existing in `cmd/cluster_stacks.go`. It returns the argument unchanged when it is 36 chars with 4 dashes, otherwise pages `ListClusters` at 100 per page up to 50 pages and matches the name with `strings.EqualFold`, erroring with `cluster %q not found` if nothing matches. So the lookup is case-insensitive and does not silently truncate at one page. `resolveClusterArg` gates it behind a stricter UUID regex first, so a malformed 36-char argument still takes the name path rather than being forwarded as an id.

Myothas:
Both fixed in 69d4edc.

**[medium] Ambiguous case-insensitive match.** Correct, and worth catching given this PR puts that resolver in front of 103 more commands including `deprovision`. An exact name now returns immediately, which keeps the early exit for the common case and honours the user's unambiguous intent. Only when no exact match exists does the case-insensitive fallback decide, and then the full listing is read before answering: one match resolves, several are refused naming both clusters, both ids, and telling the user to pass the id. Covered by `TestResolveClusterArgRefusesAnAmbiguousName` and `TestResolveClusterArgPrefersTheExactName`.

**[low] UUID-shaped cluster name.** I checked, and nothing in the CLI or the API constrains a cluster's name, so this is reachable in principle rather than ruled out. I kept the short-circuit and documented the trade-off on `resolveClusterArg`: guarding it would cost every scripted caller a listing request on every command, and the cluster id is always an unambiguous way to name the cluster. Flagging rather than fixing, since the alternative is worse for the common path.

One note on scope for the reviewer: `resolveClusterID` is pre-existing and shared with the stacks, security, cost and validate commands, so this ambiguity fix changes their behaviour too, in the same direction.

Myothas:
Two fixed in 84568e0, one answered with evidence.

**[medium] Managed clusters and the cluster listing.** They are in it. `ImportManagedClusterResponse` carries `ClusterID` and `ProviderClusterID` as two distinct fields (`internal/client/managed_clusters.go:317`): importing a provider-side cluster mints an Ankra cluster id alongside the provider's own, and `CreateManagedCluster` returns an Ankra `ClusterID` the same way. The managed lifecycle routes are addressed by that Ankra id, and there is no `ListManagedClusters` anywhere in the client, so `/api/v1/clusters` is the only inventory a managed cluster can be listed from. `DiscoverManagedClusters` is the separate provider-side inventory you were thinking of, but it enumerates clusters that have *not* been imported, and none of the widened commands can address those. So the `<cluster_id|name>` form is usable on this lane. I could not put this to a live managed cluster, since the org I checked has none, so it rests on the client contract rather than an observed call.

**[low] Looser id heuristic than isLikelyClusterID.** Good catch, and worse than low for this PR: it is the same bug the PR exists to fix, narrowed to one class of names. `resolveClusterID` now delegates to `isLikelyClusterID`, so both paths apply the strict 8-4-4-4-12 test and a 36-char, four-dash *name* goes through the listing. `TestResolveClusterArgTreatsAUUIDShapedNameAsAName` pins it, and asserts its own fixture really is 36/4 so it cannot rot into a vacuous test.

**[low] Paging cap reported as 'not found'.** Fixed. When the loop stops at the cap while the listing reports more pages, the error now says the name was not among the first 5000 and to pass the id, which is a different claim from 'it does not exist'. `TestResolveClusterArgSaysTheListingWasTruncatedRatherThanNotFound` asserts the phrase 'not found' is absent. I did not switch to the server-side name filter: `GetCluster` matches case-sensitively and would drop both the case-insensitive fallback and the ambiguity detection added in 69d4edc.

Myothas:
## Shepherd review: two surgical fixes, then a CHANGELOG rebase and it can go

The migration is mechanically uniform and genuinely well covered — `TestEveryClusterArgumentAdvertisesTheName` walks the whole command tree and reads *any* bracketed placeholder, so `[cluster_id]` / `<cluster-id>` / `<clusterID>` are all caught. That is a real ratchet. **The conflict is CHANGELOG.md only** — all 35 code/test files auto-merge, `cmd/e2e_test.go` included; master just cut v0.15.1/v0.16.0-rc0-rc2 above your `## Unreleased` block. No new `<cluster_id>` placeholder landed on master since your base, so the ratchet still passes post-merge.

Two things to fix first, both small.

### 1. `cmd/cluster.go:94-102` + `cmd/cluster_stacks.go:472` — an unresolved *name* exits 1, the same cluster as an *id* exits 3

`exitCodeFor` (`cmd/exitcodes.go:57-80`) maps a 404 `UnexpectedResponseError` to `exitNotFound` (3), and its header comment calls the codes "part of the CLI's scripting contract … treat not-found as idempotent success". `resolveClusterID` returns a plain `fmt.Errorf`, so it falls through to `exitError` (1).

This PR's whole premise is that the two spellings are interchangeable — and it makes them so on 103 commands — but they now disagree on the one thing scripts actually branch on. Concrete: a CI teardown written as `ankra cluster hetzner deprovision "$CLUSTER" || [ $? -eq 3 ]` is idempotent today with `CLUSTER=<uuid>`; the moment someone switches it to `CLUSTER=ci-ephemeral-42` (which this PR invites), a second run exits 1 and reds the pipeline.

The repo already wrote this rule down for the sibling lane — `cmd/application_resolve.go:126-130`: *"exitNotFound keeps the name path and the id path telling scripts the same thing … an id path gets a 404, which exitCodeFor already maps to exitNotFound, so a name that does not resolve must too."*

Fix: wrap **only** the genuine-absence branch (`cluster_stacks.go:472`) in `withExitCode(exitNotFound, …)`. The ambiguous (`:480`) and truncated (`:469`) branches must deliberately stay off 3 — neither is a verified absence.

### 2. `cmd/scaleway_cluster.go:259` — the only permanent-delete lane with no confirmation, now reachable by a short name

`grep -n 'confirm\|yes' origin/master:cmd/scaleway_cluster.go` returns nothing. Unlike Hetzner (`hetzner_cluster.go:121`), DigitalOcean (`:117`), OVH (`:125`) and UpCloud (`:122`), Scaleway's `deprovision` has no `confirmPrompt` and no `--yes` — while its own `Long` says "Permanently delete a Scaleway cluster and the provider resources Ankra created for it."

Until this PR, the 36-character UUID *was* the de-facto confirmation. After it: an operator with `prod` and `prod-staging` in scope types `ankra cluster scaleway deprovision prod`, resolution succeeds silently, and the cluster is gone with no prompt and nothing typed that looks dangerous.

You took care over "the eleven prompts that name the cluster" — this is the twelfth, and it is the one with no prompt at all. Give it the same `confirmPrompt` + `--yes` as the other four, using `clusterTarget(args[0], clusterID)`.

### Nits
- `resolveClusterArg` appends "(pass the cluster's name as `ankra cluster list` shows it, or its id)" to **every** resolver error, including `listing clusters: <network/5xx>`. A platform outage then reads as though the user typed the wrong thing. Append the hint only on the not-found branch.
- No memoisation: `cluster mesh readiness a b c` walks the listing once per argument. `Client.GetCluster` (`internal/client/clusters.go:88`) already does a server-side `?name=` lookup in one request; `resolveClusterID` ignores it and pages up to 50×100.
- Deprecated/Long/Example strings still spell `<cluster_id>` (e.g. `hetzner_cluster.go` `Deprecated: "use \`ankra cluster upgrade <cluster_id> <target_version>\`"`). The ratchet only reads `Use:`.
- Case-sensitivity now differs by spelling: the positional argument resolves case-insensitively, but `--cluster PROD-EU` goes through `lookupClusterByNameOrID` → `GetCluster`, an exact `cluster.Name == name` match (`clusters.go:96`).
- **One command family it misses, ratchet-invisible because it is a flag not a `Use:` placeholder**: `cmd/stack_profile_fleet.go:172 resolveClusterReference`, added to master in #285 *after* you branched. It is a verbatim copy of the pre-fix resolver — loose `len==36 && dashes==4` id test, `strings.EqualFold` first-hit, and `"cluster %q not found"` on the paging cap (`:194`). Post-merge the repo has three divergent resolvers, and `ankra stack-profiles rollout <p> --cluster prod-eu` still silently picks one of `Prod-EU`/`prod-eu`. Follow-up bead rather than growing this PR.
- `lookupClusterByNameOrID` (`cluster_select.go:203-209`) swallows the underlying error into `cluster %q not found` — the same Form-1 defect this PR removes from the positional lane. Follow-up.

### Test gaps
No test covers exit codes on any of the three failure branches — issue 1 would have been caught. Nothing proves a `ListClusters` 500 surfaces as an error rather than an absence, which is precisely the class this PR claims to fix. And nothing asserts a destructive command prompts at all, which is why the missing Scaleway prompt is invisible to the suite.

**Docs need no action:** `.github/workflows/docs-sync.yml` regenerates the ankra-docs CLI reference from the live command tree (`go run ./tools/gendocs`) on every stable `vX.Y.Z` tag, so the widened `<cluster_id|name>` usage strings reach the docs repo automatically on the next stable release.

Merge order with #280 (Scaleway deprovision dispatch): #280 first — it is 3 files and clean, this is 36 and must rebase regardless, so absorbing one CHANGELOG line is far cheaper than the reverse. Note #280's review also flags the same Scaleway deprovision lane, so issue 2 above is worth coordinating with it.

PR #280, opened 2026-09-10
fix(cluster): deprovision tears down a Scaleway cluster instead of leaking it (ankra-e3pa7)

Myothas:
Bead: ankra-e3pa7

## The defect

`cmd/reconcile.go` carried the set of "cloud" cluster kinds in **four separately hand-maintained copies** - the `cloudClusterKind` constants, `isCloudClusterKind()`, the `--force` warning switch, and the deprovision dispatch switch - plus two prose lists. **Scaleway was in none of them.**

The consequence is stated by the code's own doc comment:

> `isCloudClusterKind` reports whether deprovisioning this kind goes to the provider endpoint, which deletes the cluster record along with its resources. The generic imported lane keeps the record, so the two are not interchangeable and the difference has to reach the operator.

So `ankra cluster deprovision <a-scaleway-cluster>` did not recognise it as a cloud kind, fell through to the generic imported lane, and **never called the Scaleway endpoint**. The CLI printed that the deprovision had started. The cluster record was handled as an import. The instances, private network, security group, public gateway and private NICs kept running and kept billing, with nothing telling the operator the teardown had not happened.

`DeprovisionScalewayCluster` already existed on both `internal/client` and the `APIClient` interface. It was simply never wired up - so this is missing wiring, not a missing feature.

Nothing caught it because every sibling test enumerates providers by hand and omitted Scaleway in exactly the same way (`cluster_test.go`, `cluster_scale_test.go`, `cluster_ssh_keys_test.go`, `cluster_upgrade_test.go`).

## The fix

- `allCloudClusterKinds` is now the one list the decisions derive from, and `isCloudClusterKind` derives from it rather than restating it.
- Scaleway gets its dispatch case, calling `DeprovisionScalewayCluster`.
- Both prose lists updated.

Scaleway's endpoint takes no `force`, like Proxmox and Morpheus, so it correctly keeps falling into the existing "`--force` has no effect for this cluster type" branch.

I did **not** claim the list makes the dispatch a compile error - Go does not check switch exhaustiveness. The comment says what the list actually buys: one place for the tests to enumerate.

## Closing the class

`TestEveryCloudClusterKindHasADeprovisionDispatch` walks `allCloudClusterKinds` and, for each, drives the real command through a mock, asserting the **provider** endpoint was called and the **generic** lane was not. A provider added to the list and nowhere else fails here instead of silently leaking a customer's infrastructure.

`TestIsCloudClusterKindCoversEveryCloudKind` pins the record-deletion warning against the same list, and pins that an unknown or empty kind does *not* take the cloud lane.

**The guard is verified against the defect**, not just against the fix: removing only the Scaleway dispatch case (leaving the kind in the list) makes it fail with

```
a scaleway cluster was deprovisioned through the GENERIC imported lane.
```

A guard that passed on both broken and fixed code would be worthless, so this was checked rather than assumed.

## Verification

- `go build ./...` clean; `gofmt` clean on both changed files
- Full suite: 6 packages ok, no failures
- New tests: 7 dispatch subtests + the coverage test, all passing
- Guard proven to fail on the reintroduced defect

`gofmt -l` also reports `cmd/cluster_mesh_test.go`, which is pre-existing on `master` and not touched here - left alone rather than folded in as an unrelated reformat.

## Operational follow-up, not fixed by this PR

Any Scaleway cluster already "deprovisioned" through this command still has its infrastructure running and billing in the customer's Scaleway account. The code fix stops new occurrences; it cannot release what was never released. Someone needs to check for orphaned Scaleway resources and tell affected customers. Tracked on the bead, and flagged in the CHANGELOG entry so operators see it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## Unreleased
 
+### Fixed
+
+- **`ankra cluster deprovision` now actually tears down a Scaleway cluster.**
+  Scaleway was missing from every cloud-provider list behind the command, so
+  a Scaleway cluster fell through to the generic imported lane: the CLI
+  reported the deprovision had started, the cluster record was handled as an
+  import, and the instances, private network, security group, public gateway
+  and private NICs were never released - they kept running and kept billing,
+  with nothing saying the teardown had not happened. The command now calls the
+  Scaleway endpoint, and warns as it does for the other cloud providers that
+  the cluster record itself is deleted and cannot be provisioned again. If you
+  ran `cluster deprovision` against a Scaleway cluster before this release,
+  check your Scaleway account for resources still running.
+
 - **A Claude Design export becomes a deployable application in one command.** `ankra application import claude-design <path>` takes what you exported from claude.ai/design - a directory of `<Name>.dc.html` artboards with `canvas.json` and images, a zip of it, one artboard, or a saved canvas page - and has Ankra convert it into a static site, create a GitHub repository under your GitHub credential (`--credential`, `--owner`), commit it with a Dockerfile and register the application, so the setup pull request, build and deploy follow as for any other application. `--repository`, `--visibility` and `--source-url` shape the repository; `--wait` follows the analysis to the setup pull request; `-o json` returns the pages and any warnings. Artboards that depend on the Claude Design runtime are kept as authored and reported, and re-running the same import registers the same repository without a second commit. The lane ships behind the `claude_design_import` organisation feature flag: while it is off the command explains that and exits 3, so ask Ankra support to enable it for your organisation.
 
 ## v0.15.1 — 2026-09-09
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,17 +4,20 @@
 
 ### Fixed
 
-- **`ankra cluster deprovision` now actually tears down a Scaleway cluster.**
-  Scaleway was missing from every cloud-provider list behind the command, so
-  a Scaleway cluster fell through to the generic imported lane: the CLI
-  reported the deprovision had started, the cluster record was handled as an
-  import, and the instances, private network, security group, public gateway
-  and private NICs were never released - they kept running and kept billing,
-  with nothing saying the teardown had not happened. The command now calls the
-  Scaleway endpoint, and warns as it does for the other cloud providers that
-  the cluster record itself is deleted and cannot be provisioned again. If you
-  ran `cluster deprovision` against a Scaleway cluster before this release,
-  check your Scaleway account for resources still running.
+- **`ankra cluster deprovision` now works on a Scaleway cluster.** Scaleway was
+  missing from every cloud-provider list behind the command, so a Scaleway
+  cluster fell through to the generic imported lane, which the platform
+  refuses for cloud clusters: the command printed `Deprovisioning cluster: X`
+  and then failed with `409: This cluster is a scaleway cloud cluster. Use
+  DELETE /api/v1/clusters/scaleway/{cluster_id}...`, and you had to fall back
+  to `ankra scaleway cluster deprovision <id>`. The confirmation prompt also
+  omitted the warning the other providers show, that the cluster record itself
+  is deleted and cannot be provisioned again. Both are fixed. Nothing was
+  silently left running: the platform has refused that route for Scaleway
+  since 2026-08-19, so the teardown either ran through the provider command or
+  visibly failed.
+
+### Added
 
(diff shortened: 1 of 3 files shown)


Myothas:
## Shepherd review: code is right, the shipped narrative is not

The dispatch change is correct, minimal and a line-for-line twin of the Proxmox/Morpheus cases, and it correctly stays out of the `--force` branch since `DeprovisionScalewayCluster` hardcodes `force=false`. Two things to fix before this merges.

### 1. `CHANGELOG.md:7-17` asserts a silent infrastructure leak that did not happen

`cluster/go/internal/usecase/providers/lifecycle.go:403` refuses the generic deprovision route for Scaleway: `containsState([]string{Hetzner, Ovh, Upcloud, Digitalocean, Proxmox, Morpheus, Scaleway}, clusterKind)` → `LifecycleInvalidState` → `cliapi/lifecycle.go` `writeDetail(writer, http.StatusConflict, ...)`, and the CLI treats non-200 as an error. That refusal was added deliberately in cluster `4361e2ccf7` ("Refuse the generic imported-deprovision route for Scaleway clusters", **2026-08-19**).

So on every platform since then the command printed `Deprovisioning cluster: X` and then **failed loudly** with `409: This cluster is a scaleway cloud cluster. Use DELETE /api/v1/clusters/scaleway/{cluster_id}...`. It never reported success and never left anything running that the operator was not told about.

As written, this ships a changelog telling every customer *"If you ran `cluster deprovision` against a Scaleway cluster before this release, check your Scaleway account for resources still running"*, and the bead dispatches someone to contact customers about a leak that did not occur.

Fix: state the real defect — the command 409'd and you had to fall back to `ankra scaleway cluster deprovision <id>`, and the confirm prompt omitted the "record is DELETED" warning — then either drop the audit instruction or scope it to platforms older than 2026-08-19. The in-code comment (`cmd/reconcile.go` ~324-336) and the test's failure message (`cmd/cloud_deprovision_dispatch_test.go:12-19`) carry the same wrong mechanism and need the same correction.

### 2. The class is not closed — the real silent leak is one kind-family over, and the new guard structurally cannot see it

The 409 refusal covers only the 7 **self-hosted** kinds. The 7 **managed** kinds (`kapsule, doks, uks, gke, ovh_mks, aks, eks` — `clusterrevoke.go:125`) are refused only by the imported *delete* route (`importedwrite/lifecycleops.go:51`), **not** by `DeprovisionImportedCluster`.

So today: `ankra cluster deprovision <a-doks-or-kapsule-cluster>` → kind not in `allCloudClusterKinds` → generic lane → HTTP 200 with `marked_for_deprovision_at` → CLI prints "Cluster deprovision initiated." → backend sets all resources to stopping and reconciles, while the managed control plane and node pools are never deleted and keep billing. That is verbatim the bug this PR claims to have fixed, still live, for 7 kinds.

`TestEveryCloudClusterKindHasADeprovisionDispatch` iterates the CLI's own list, so a kind missing from that list is invisible to it — the guard cannot fail on the case that actually leaks. Please file a P1 bead for the managed-kind lane and soften the claim that `allCloudClusterKinds` is "the one list the cloud-kind decisions derive from" (it is one of two, and the wrong one for managed kinds). Do **not** just add the managed kinds to this list — they need `cluster managed delete` routing, which this dispatch does not have.

### Nits
- `CHANGELOG.md:5`: the new `### Fixed` heading sits directly under `## Unreleased`, so the pre-existing **Claude Design import** bullet (a feature) is now published under "Fixed". It needs an `### Added` heading above it.
- Test failure message renders `case cloudClusterKindscaleway:` (lowercase `%s` of the kind, not the const name).
- The dispatch test never asserts *which id* reached the provider — passing `clusterName` instead of `clusterID` to a DELETE would still pass. One `if gotClusterID != "c-1"` pins the wrong-target axis on a destructive path.
- `--force`'s honoring set (`hetzner, upcloud, ovh, digitalocean`) is still a second hand-written list derived from nothing, untested per kind — the same class this PR set out to close.
- `result.Success` is never inspected in any of the 7 branches (pre-existing): a 200 with `success:false` prints "deprovision initiated".
- `resolveClusterFromArgsWithKind` returns kind `""` when the cluster lookup **fails**, routing an UNKNOWN kind to the generic lane — and `TestIsCloudClusterKindCoversEveryCloudKind` now *pins* `isCloudClusterKind("") == false` as intended. Harmless today because the backend is authoritative and 409s, but it writes our named UNKNOWN→wrong-lane collapse into a test rather than fixing it.

Merge order note: only `CHANGELOG.md` overlaps with #275 (both insert at the top of `## Unreleased`). #275 does not touch `cmd/reconcile.go`. **This one should go first** — it is 3 files and clean, whereas #275 is 36 files and already conflicting, so it must rebase regardless.

PR #291, opened 2026-09-11
feat(pipeline): wait on a run you did not dispatch, and carry its outcome into the exit code (ankra-gxs9p)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-843/1165-cli-cannot-wait-on-or-watch-a-pipeline-run-you-did-not-dispatch
Bead: ankra-gxs9p

## What this is

Smartoptics moved two repositories fully onto Ankra Pipelines and deleted their GitHub
Actions workflows, so Ankra is now the only build and deploy signal for them. They then
found that the single most common thing their automation does — wait for a run and branch
on whether it passed — is the one thing the CLI could not do.

This ships asks 1 and 3 of the five in PLA-843. Asks 2, 4 and 5 are filed as beads
(below); none of them blocks this.

## The gap

`--wait` existed only on `pipeline run` and `pipeline rerun`, and both of those *dispatch*
the run they then wait on. Every push and pull_request run is dispatched by the webhook
instead, so for the majority of runs there was nothing to wait with: `pipeline get` had no
`--wait`, and nothing under `pipeline` had `--watch`. Consumers were writing a bash loop
around `pipeline list` that polls every 60 seconds and diffs the status string.

`runPipelineGet` also returned `nil` unconditionally, so a script could not branch on exit
status and had to parse `outcome` back out of the payload.

## The bug found on the way

`pipeline run --wait -o json` and `pipeline rerun --wait -o json` exited **0 on a failed
run**. `pipelineRunConclusionError` exists precisely to make a non-success conclusion a
non-zero exit, and the human-table branch reached it — but the structured branch returned
`encodeStructured`'s result and never got there:

```go
if format != outputDefault {
    return encodeStructured(command.OutOrStdout(), format, detail)   // <- returns nil
}
printPipelineRunDetail(command.OutOrStdout(), *detail)
return pipelineRunConclusionError(detail.PipelineRun)
```

`--wait -o json` is exactly the shape a CI script uses, so a failed build could pass a
pipeline step silently. This was not in the customer's report; it is the same defect class
as their ask 3 and is fixed here. Rendering and the verdict are now two separate steps in
one shared `waitForAndReportPipelineRun`, so the two output formats cannot drift again.

## What changed

- **`pipeline get --wait`** blocks until the run concludes, announces each status change on
  stderr, prints the final detail on stdout, and reports the outcome through the exit code.
  It reuses the existing `waitForPipelineRunConclusion` and `pipelineRunConclusionError`.
- **`--timeout` on `get`/`run`/`rerun`**, defaulting to **0 = no limit**. That default is
  deliberate: waiting indefinitely and giving up on Ctrl+C is the contract `run --wait`
  shipped with and its doc comment defends, so nothing an existing invocation means
  changes. An expired budget exits `5` (`exitWaitTimeout`) and says the run keeps going. A
  negative duration is a usage error, not a silent "no limit".
- **`pipeline get --exit-code`** reports a *concluded* run's outcome as the exit status
  without waiting. A run still going exits 0 under it — "has not concluded" is not "did not
  succeed", and `--wait` is how you ask for a verdict.
- **A bare `pipeline get` is unchanged** and still exits 0 whatever it finds. The customer
  explicitly offered this as the compatibility-safe option and it is the one taken: every
  script that prints a run's detail today keeps working.
- Every flag is registered on the `ankra application pipeline …` twin too, with a test that
  fails if the two surfaces drift.
- CHANGELOG, and the `ankra-cicd` embedded skill gains the waiting recipe.

## Verification

- `go build ./...`, `go vet ./cmd/...` clean; `gofmt -l` reports only two files this branch
  does not touch (`internal/client/baseurl.go`, `internal/client/helpers_test.go`, both
  unformatted on `master`).
- `go test -count=1 ./...` — see the checks on this PR. `golangci-lint` is not installed
  locally; the full linter runs in CI.
- New tests in `cmd/pipeline_wait_test.go` cover: `get --wait` polling to conclusion; a
  failed run exiting non-zero under `--wait`; the `-o json` regression on all three of
  `get`/`run`/`rerun` (asserting the payload is *still* printed); a bare `get` staying exit
  0; `--exit-code` on a concluded run and its silence on a running one; `--timeout` expiry
  exiting 5; a negative `--timeout` refused; and both command surfaces carrying the flags.

## Not in this PR

Filed as beads, each with the design question that needs a human:

- `ankra-0yxgz` — ask 2, addressing a run by `--head-sha`/`--trigger`/`--latest` instead of
  a run id. The listing filters already exist; what needs deciding is how `--latest`
  breaks ties and what "no run for this commit yet" exits as, since a webhook run lags the
  push by seconds.
- `ankra-57d0z` — ask 4, `--watch` emitting NDJSON per state transition. Nothing streams
  *run* state today (only step logs), so this is either a new platform stream or a
  documented client-side diff over the same poll, and that choice sets the latency promise.
  It should also agree on an envelope with `ankra-y638.7.2`.
- `ankra-g3ra5` — ask 5, `logs` on a concluded-but-archiving step. The reported symptom
  ("blocks, still open at a 120s timeout") **does not reproduce**: that branch prints to
  stderr and returns immediately, and `v0.15.2` is byte-identical to `master` there. The
  real defect is that it returns exit 0 with empty stdout, which is indistinguishable from
  a step that printed nothing — which is why a retry wrapper looked like a hang. Fixing it
  wants an exit-code decision: 8/9/10 are already reserved by `ankra-y638.7.1`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,38 @@
 
 ## Unreleased
 
+### Added
+
+- **`ankra pipeline get --wait` waits for a run you did not dispatch.** `run
+  --wait` and `rerun --wait` can only block on a run they started themselves,
+  and every push and pull_request run is started by the webhook instead - so
+  the one run a CI script most wants to wait for was the one nothing could
+  wait for, and consumers were writing their own poll loops around `pipeline
+  list`. `get --wait` blocks until the run concludes, announces each status
+  change on stderr, prints the final detail, and reports the outcome through
+  the exit code. Available on `ankra application pipeline get` too.
+- **`--timeout` bounds any `--wait` on the pipeline lane** (`get`, `run`,
+  `rerun`). It defaults to no limit, which is the contract `run --wait`
+  already had - a caller who wants to give up presses Ctrl+C - and an expired
+  budget exits 5, the scripting contract's wait-timeout code, with a message
+  saying the run keeps going and how to follow it. A negative duration is
+  refused as a usage error rather than read as "no limit".
+- **`ankra pipeline get --exit-code` reports a concluded run's outcome as the
+  exit status** without waiting, for a script that already knows the run has
+  settled. A bare `get` is unchanged and still exits 0 whatever it finds, so
+  nothing that prints a run's detail today changes behaviour. A run that has
+  not concluded exits 0 under `--exit-code`: "still going" is not "did not
+  succeed", and `--wait` is how you ask for a verdict.
+
 ### Fixed
+
+- **`--wait -o json` no longer exits 0 on a failed run.** `pipeline run
+  --wait` and `pipeline rerun --wait` reported a non-success conclusion
+  through the exit code only when rendering the human table; the structured
+  branch returned as soon as it had encoded the payload and never reached the
+  verdict. That is precisely the invocation a CI script uses, so a failed
+  build could pass a pipeline step silently. The payload is still printed;
+  the exit code is now the same answer in both formats.
 
 - **Every cluster-scoped command now takes the cluster's name as well as its
   id.** The three `ankra cluster playground` verbs learned this in #239, but
--- a/cmd/application_pipeline.go
+++ b/cmd/application_pipeline.go
@@ -86,6 +86,7 @@ func newApplicationPipelineGetCommand() *cobra.Command {
 			return runPipelineGet(command, client.PipelineSelector{ApplicationID: applicationID}, arguments[1])
 		},
 	}
+	registerPipelineGetFlags(getCommand)
 	registerStructuredOutputFlags(getCommand)
 	return getCommand
 }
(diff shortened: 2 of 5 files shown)


Myothas:
Ankra AI agent (stacks coordinator, fleet round r-20260911T205730Z-7b7467). **Closing as superseded by #292** (twin-PR ruling for PLA-843 / ticket 1165):

- #292 is a scope superset: all five asks in order (this PR ships asks 1 and 3) and it carries the same `run --wait -o json` exit-0 fix.
- This PR cites `Bead: ankra-gxs9p`, which does not exist in the ankra tracker; `ankra-w1i58` is the tracker's record for PLA-843 and #292 cites it.
- Verified overlap on CHANGELOG.md, cmd/application_pipeline.go, cmd/pipeline_run.go and cmd/pipeline_wait_test.go: both cannot land.

The one thing this branch has that #292 lacks, the `ankra-cicd` embedded skill recipe, is asked of #292 in a comment there. Nothing is deleted; reopen if the ruling is wrong.

PR #293, opened 2026-09-11
feat(aws): ankra cluster aws commands for self-managed k3s on EC2 (ankra-rtpno.13)

Myothas:
## Summary

CLI surface for the AWS self-managed k3s-on-EC2 lane (closed beta behind `aws_provider`).

- `ankra cluster aws` — create, preflight, deprovision, stop, start, workers, k8s-version, access-info, regions, instance-types, vpcs, subnets, availability-zones, images, pricing, control-plane, nodes (list/restart/cloud-init-log), bastion (status/diagnose).
- `ankra credentials aws` — list (with each credential's auth method and onboarding scope), onboarding, create-role, create-keys (secret via masked prompt, never a flag).
- Request/response types mirror the cluster OpenAPI schemas exactly; unknown prices render as `-`, never `0`.
- `aws` joins `allCloudClusterKinds` with a deprovision dispatch and every hand-maintained provider switch; Scaleway was added to the two test tables it was missing from.
- System test lane (`systemtest/lifecycle_systemtest.sh` + workflow): opt-in when `SYSTEMTEST_AWS_*` secrets exist — preflight, create, access-info, nodes, stop/start, deprovision, then a tag-based leak check (instances, security groups, key pairs, EIPs, route tables, volumes, IAM roles/profiles) and a customer-VPC-untouched diff of route tables, subnets and DHCP options.

Beads: ankra-rtpno.13 ankra-rtpno.17

Depends on ankraio/cluster#2955 merging first: the route census (`TestClusterRoutesAreRegistered`) passes against that branch's `routes.json` with no allowlist change and will fail against `main` until then.

## Verification

`go build ./...`, `go test -race -count=1 ./...` green (golang container); route census green against the cluster branch; shellcheck clean on the system-test script.

## Live run

This CLI binary drove a real cluster-api + scheduler + cluster-maintenance stack (Postgres, NATS, Vault, moto as the AWS API double) through flag gating, every catalog, onboarding and scope validation, request validation, preflight (bastion_nat, existing, flannel warning), create, reads, the scheduler's create lane up to the bastion SSH wait, deprovision with the customer VPC untouched, and the quotas/VMs refusals. The one CLI defect it found (scope and auth_method dropped from `credentials aws list`) is fixed here. Static analysis: 0 issues.


CodeStaple:
Triage of the Ankra AI review (at c600a35), fixes in 44550f4:

- **[medium] Zero-valued int flags: not a bug.** `ControlPlaneCount`, `EtcdNodeCount` and `RootVolumeGiB` are tagged `omitempty` in `internal/client/aws_clusters.go` (lines 49, 57, 64), and 0 is never a valid value for any of them, so an unset flag is omitted and the server default applies. `WorkerCount` is a pointer only because 0 workers is legitimate (node-group-only clusters).
- **[medium] `stop --force` without confirmation: consistent with the fleet, tracked fleet-wide.** No provider prompts on `stop --force` today: `hetznerStopCmd`, `ovhStopCmd`, `upcloudStopCmd` and `digitaloceanStopCmd` read `--force` and call the API directly; only `deprovision` prompts. Adding a prompt to AWS alone would make one provider's scripts behave differently, so it is filed as ankra-ovzub to land on every provider in one change (prompt only with `--force`, `--yes` escape, exit 4 on decline).
- **[low] `AWS_REGION` defaulted despite the header: fixed.** The `eu-west-1` fallback is gone and `SYSTEMTEST_AWS_REGION` is now part of the opt-in gate, so the lane is never added against a guessed region.
- **[low] `workflow_dispatch` input expanded inside `run`: fixed.** The providers input and the `aws_configured` output now reach the script as `PROVIDERS_INPUT` / `AWS_CONFIGURED` through `env`. actionlint clean.

The red `lint-test-build` is `TestClusterRoutesAreRegistered`, which checks against cluster `main`; it goes green once cluster#2955's `/api/v1/clusters/aws` routes land.

PR #295, opened 2026-09-12
feat application: say which registry an added application will publish to

CodeStaple:
`ankra application add` with no `--registry-url` points the application at the organisation's own Ankra registry project and says nothing about it. The `Registry:` line was printed only for a declaration the user had made themselves — so the one case where the user does not already know the answer is the one case nothing was reported.

That is PLA-825, from Smartoptics, who run their own Harbor alongside the Ankra registry. The reporter narrowed it themselves: **discovery is not missing, the default is just unstated.** It is a defensible default and an invisible one, and `--registry-url`'s own help already warns that declaring a registry later leaves a workflow logging in to the wrong one — so the moment to say it is at the end of `add`, while the analysis is still running and a correction is still cheap.

## What it prints now

```
  Registry:   registry.ankra.cloud (Ankra's own, the default - no --registry-url was given)
```

and, only when it has something to say:

```
Other applications in this organisation publish to artifact.smartoptics.dev.
If this one should too, declare it now, before the setup pull request is generated:
  ankra application registry set <id> --url oci://<host>/<project> --credential <name>
The build workflow is generated from the declaration the application is created with,
so a registry added after that leaves a workflow logging in to the wrong one.
```

That second half is what catches the real mistake. An organisation with four applications on `artifact.example.com` and a fifth silently on Ankra's registry is not expressing a preference — it is missing a flag.

## Best-effort by construction

The sibling listing runs **after** the application exists, and every failure answers nil: a bad listing, an unparseable payload, no permission. A hint is not worth an error on a command that has already succeeded, and a user who cannot list applications is not a user this sentence helps.

It reads one page of at most 100. An organisation whose first page is entirely on the Ankra registry is not one where the answer changes on page two, and `add` must not become slow to tell the user something it could have left unsaid.

## Tests

`TestApplicationAddCommandOmitsTheRegistryWhenUndeclared` asserted the silence this removes; it now asserts the notice — the default is named, the Ankra host is named, and the overriding flag is named. Its other half, that `image_registry` is omitted from the create request entirely rather than sent empty, is untouched and is still the invariant that matters about the request itself.

New `TestRegistryHostOfImageURL` covers `oci://`, `https://`, `http://`, bare host, host:port, a repository path, whitespace and empty. `TestAnkraRegistryHostIsTheStoredSpelling` pins the constant against the platform's own spelling, since a drift in either would silently stop the hint telling Ankra's registry apart from one the organisation operates.

Full `ankra/cmd` suite green; pre-commit ran the whole test set and the linter with 0 issues.

Bead: ankra-0uxap

### DIFF
--- a/cmd/application.go
+++ b/cmd/application.go
@@ -60,11 +60,14 @@ remote's default branch when available, and falls back to the current branch.
 It selects an available GitHub credential automatically when the choice is
 unambiguous.
 
-Pass --registry-url to have the application publish to a container image
-registry you already operate instead of the organisation's own Ankra registry
-project. Declare it here rather than afterwards: the setup job generates the
-build workflow from the declaration the application is created with, so a
-registry added later leaves a workflow that logs in with the wrong one.`,
+Without --registry-url the application publishes to the organisation's own
+Ankra registry project, and the command says so on success. Pass it to publish
+to a container image registry you already operate instead. Declare it here
+rather than afterwards: the setup job generates the build workflow from the
+declaration the application is created with, so a registry added later leaves
+a workflow that logs in with the wrong one - which is also why, when other
+applications in the organisation publish somewhere else, this command names
+those registries before the analysis finishes.`,
 		Example: `  ankra application add .
   ankra application add ./services/payments --name payments
   ankra application add . --credential github-acme --branch main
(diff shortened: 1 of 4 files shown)


CodeStaple:
Both substantive findings land — fixed in `82c7da3`.

**Hardcoded host (medium).** You are right, and it is worse than a style point: printing `registry.ankra.cloud` as the application's registry asserts as fact something the CLI was never told. I checked whether the effective value could be read from the create response instead — it cannot: `CreateApplicationResponse` carries `id` and `errors` and nothing else. So the notice now names the registry by what it *is*, "the organisation's own Ankra registry project", and leaves the host to the surfaces that actually read it back (`ankra application list` reports `container_image_url` per application). The constant stays, because comparing against it is still how a sibling registry is told apart from Ankra's own, and its doc comment now says printing it is not allowed. A test asserts the host does **not** appear in the output.

**Contradictory advice (low).** Also right. The remedy is the flag, so the flag is what it names now — "add it with `--registry-url` instead" — followed by the reason. One instruction, consistent with the sentence under it, and nobody is walked into the state the warning describes.

**`ListApplicationsRaw` (low).** It exists: `internal/client/application_resources.go:175`, signature `(ctx, page, pageSize, search)`. The package compiles and the full `ankra/cmd` suite passes; pre-commit runs both plus the linter with 0 issues.

PR #298, opened 2026-09-12
fix application build: attribute build_unknown ankra-8pm5h

CodeStaple:
Bead: ankra-8pm5h

## What

`ankra application build` now knows the `build_unknown` build error class.

- `describeFailedBuild` has a `build_unknown` arm. Before, such a build fell through to the default arm and printed a bare "the build failed". That reads as the repository's fault. It now says the build ran and failed unclassified, and that its message is the only copy of the build output Ankra keeps, so the user should include it when reporting. A build with no recorded output says so.
- The `application build get` help text lists `build_unknown` with the other classes: Ankra's, but not already visible to Ankra the way `capacity` is.
- CHANGELOG `## Unreleased` gets a Fixed entry.

## Why

cluster#2961 split the builder's catch-all out of `capacity` into `build_unknown`, and cluster#2970 put it live on the builders. The producer's wording, which this mirrors (`scheduler/go/cmd/platform-builder-agent/classify.go`, `enginekit/platformbuild`):

- **`capacity`** means nothing ran.
- **`build_unknown`** means a builder claimed the build, ran `buildctl`, and it failed matching none of the classifier's probes. The `error_message` holds the last 1500 bytes of output and is the only copy.

The platformbuild doc also calls out this CLI's old help text: "already visible to Ankra without anyone reporting them" is true of a dead fleet, false of a build whose output nobody kept.

## Pipeline step classes

cluster#2965 added the step class `registry_push_failed`. ankra-cli does not enumerate pipeline step error classes anywhere:

- `pipeline get`, `list`, `wait` and `logs` render a step's outcome, exit code and `error_message`.
- `PipelineStep.ErrorClass` is only a passthrough wire field, shown in `-o json`.

So no change is needed for it here. Its outcome stays `infra_error`, which the CLI already renders.

## Tests

- `go test ./cmd/ -run 'TestDescribeFailedBuildAttributesEachErrorClass|TestApplicationBuild'`: new table test over every build error class, including `build_unknown` with and without output and an unrecognised class.
- `go build ./...`
- `golangci-lint run ./cmd/...`: 0 issues.
- The pre-commit hook ran `go test ./...` and lint.

Users get this in the next CLI release. No release is cut here.


CodeStaple:
Re the one open question in the review, whether `errors` is imported: it is. `cmd/application_build.go` already imports `errors`, and its existing `build_failed` arm uses it (`errors.New("the build failed: the recipe did not build")`); `cmd/application_build_test.go` imports it too. The diff only shows new lines, so the import is out of view. `go build ./...`, `golangci-lint run` and the `lint-test-build` check all pass on fdb0ec1.

CodeStaple:
Same question re-posted for 45ad6ea, same answer: `errors` is already imported in `cmd/application_build.go` and used by the pre-existing `build_failed` arm (`errors.New("the build failed: the recipe did not build")`). The review sees only the diff, so the import line is out of view. 45ad6ea only merges master's v0.16.1 changelog section; `lint-test-build` passed on it, and so did the pre-commit `go test ./...` and golangci-lint.

PR #299, opened 2026-09-13
feat org ci-settings get shows the platform builds grant ankra-vn0bd.3.10.2

CodeStaple:
Bead: ankra-vn0bd.3.10.2

## What & why

`ankra org ci-settings get` could only print a caveat that the platform-builders capability might be off. The platform had no way to report that capability: the organisation flag `platform_builds`, which gates Ankra's own build lane. Smartoptics hit exactly this on PLA-850: the command read `platform_builders` with no hint whether the capability behind it had been granted.

ankraio/cluster#2984 (https://github.com/ankraio/cluster/pull/2984) adds a read-only `platform_builds_enabled` to `GET /api/v1/org/ci-settings`. This PR shows it:

- **Field true:** prints `Platform builds enabled: yes` and drops the caveat.
- **Field false:** prints `Platform builds enabled: no`. When the fallback is `platform_builders`, it adds a note: Ankra has not granted the capability, so a build step that fails with "build fallback is 'none'" is missing that grant, not the setting, and only Ankra can enable it.
- **Field absent** (a platform older than cluster#2984): the grant is unknown, not false. No line is printed for it, `-o json` leaves the key out, and the previous caveat stays. The caveat now names the stopgap: `ankra application build list <application-id>` answering 404 means the capability is off.
- `-o json` / `-o yaml` carry `platform_builds_enabled` as the platform reported it. The client field is `*bool` with `omitempty`, so absence is never invented as `false`.

The embedded `ankra-cicd` skill paragraph and `CHANGELOG.md` (`## Unreleased`) are updated to match.

## Test plan

- New tests in `cmd/org_ci_settings_test.go`:
  - field true: the grant is shown and no note is printed
  - field false: the denial is shown, the note says whose grant is missing, and the old caveat does not print
  - field absent: the caveat stays, names the stopgap, and prints no grant line
  - `-o json` for granted, denied and unreported: true, false, and key absent
- The existing caveat tests (fallback `platform_builders` warns, fallback `none` stays quiet) still pass unchanged.
- Pre-commit hook: `go test ./...` passes and `golangci-lint run` reports 0 issues.

## Docs checklist

- [x] No user-facing command/flag changes — no docs needed
- [ ] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); ensure `Short`/`Long`/`Example` on new commands are written for docs readers

It reaches users only in the next CLI release.


CodeStaple:
On the open question about structured output: the note cannot leak into `-o json` / `-o yaml`. Both `get` and `set` call `renderStructured` first and return when it renders (`if rendered, renderError := renderStructured(cmd, settings); rendered || renderError != nil { return renderError }`). Only the human path reaches `renderOrganisationCISettings`, and `renderPlatformBuildsNote` is called from inside that function. `TestRunOrgCISettingsGet_StructuredOutputCarriesThePlatformBuildsGrant` pins it: in the granted, denied and unreported cases it decodes all of stdout as JSON, and the decode would fail if a note line were printed. The unreported case is the one that still prints the caveat on the human path. The head the review read, `905687f`, is CodeStaple's merge of `origin/master` into the branch; it adds no change to this surface.

PR #303, opened 2026-09-13
fix pipeline: name a failed run error class in pipeline get ankra-edt1b PLA-851

CodeStaple:
## What

`ankra pipeline get <run>` now prints a failed run's **error class**, and `--wait` / `--exit-code` carry it in the error they exit with.

```
  Status:    failure
  Class:     registry_push_failed
  Error:     The build ran, and its image could not be pushed to the image registry it
             publishes to; its builder reported "push_failed". BuildKit stopped with:
             failed to solve: failed to push artifact.example.dev/smart-hub/backend:sha-abc1234: 401 Unauthorized
```

```
$ ankra pipeline get <run> --wait
Error: run #98 concluded infra_error (registry_push_failed): The build ran, and its image ...
```

## Why

`client.PipelineRun.ErrorClass` has been on the wire since this command shipped, and only `-o json` ever showed it. `printPipelineRunDetail` printed the message alone, so a run read the way runs are actually read never said whose failure it was.

That is the expensive half of **PLA-851** (Smartoptics, Ankra ticket #1173). A platform-builders build whose push their Harbor refused concluded with a fragment of the BuildKit transcript in `error_message`, and nothing on the run said `registry_push_failed`. They audited five Dockerfiles across two repositories - images that had built unchanged for weeks - before finding a 401 from their own registry. Their words: *"or at minimum surface error_class on the run itself"*.

The message half already shipped platform-side (cluster#2983, #2985, #2986 - the class sentence plus buildctl's own fatal line with the registry host and status, live since 2026-09-13T00:38Z). This is the class half, on the surface the customer actually read.

Verified on the wire before and after, against ankra-ab `conductor-hub` run #9 (`64ead339`), whose recorded class is `image_gate_blocked`: absent from the default output on `master`, printed on this branch.

## Design notes

- **Verbatim, not mapped to prose.** The message is already the sentence; the vocabulary grows server-side (`build_runtime_confined`, `build_fallback_unsupported`, `registry_push_failed` all arrived after this command shipped), and a mapper that has not been taught a new class renders it as nothing at all. An unrecognised class is still a search term; a blank is not. Human labels for the portal are tracked separately (ankra-mc9bx, ankra-q858c).
- **No class recorded prints no line**, rather than a class called `""` - a successful run, and a run that failed before anything classified it, are both "not recorded". A whitespace-only class is treated the same.
- **The message is not re-indented.** A platform build's message carries the tail of the build transcript whole and deliberately (cluster's `platformBuildClassMessage`); Ankra keeps no other copy of it, so lining it up under a label would corrupt the evidence.
- Outcome and class stay distinct in the `--wait` sentence: `infra_error` is the pipeline's word, `registry_push_failed` is the step's.

## Testing

- `go test -race -count=1 ./...` green; `make lint` 0 issues.
- Three new tests in `cmd/pipeline_test.go`: the push case renders `Class:` above `Error:` and keeps the registry host and 401 in the message; an unrecorded or whitespace-only class prints no line and leaves the message rendering unchanged; the `--wait` conclusion error names both outcome and class, and adds nothing when no class was recorded.

Bead: ankra-edt1b

@ankraai

https://claude.ai/code/session_01TcQWijf5CCcLW1jognWo64


CodeStaple:
Both review findings are compile-time questions, and both are confirmed resolved.

- **`strings` import**: already present at `cmd/pipeline_run.go:15`. The file used `strings.TrimSpace` before this change (`pipelineRunConclusionError` reads the outcome that way), so nothing was added to the import block - which is why the diff does not show it.
- **`client.PipelineRun.ErrorClass`**: `*string`, declared at `internal/client/pipelines.go:88` (and the same field on `PipelineStep` at :140). It has been on the wire since the type was written; this PR is only about printing it.

Verified locally before pushing: `go build ./...`, `go test -race -count=1 ./...` (all packages ok) and `make lint` (0 issues) are green, and the pre-commit hook ran the same test and lint gate on the commit. A missing import or a wrong field type would have failed the build on the first of those.

Also verified against production rather than only in tests: built this branch and ran it against ankra-ab `conductor-hub` run #9 (`64ead339`), whose recorded class is `image_gate_blocked`.

    Status:    failure
    Class:     image_gate_blocked
    Error:     The image gate blocked this run on 5 finding(s): ...

On `master` the `Class:` line is absent and the class is reachable only through `-o json`.

PR #305, opened 2026-09-13
fix cluster node-group list: show joined nodes next to count and name worker records with no node (ankra-a56o8.1)

CodeStaple:
Linear: https://linear.app/ankra/issue/PLA-853/1175-node-group-reports-count3-while-the-cluster-has-2-workers-and
Bead: ankra-a56o8.1

Lands after ankraio/cluster#3028, which serves the two fields this reads. Against a platform without them the CLI renders exactly as today, so the order only decides when the column appears.

## Problem

Every count the platform reports comes from its worker records, and a record can stay on the books with no Kubernetes node behind it. The customer's `ankra cluster node-group list` said `count=3` while kubectl showed two workers, and the only place the gap was visible was the failed execution steps.

## Change

`ankra cluster node-group list` (and the provider-specific `cluster hetzner|ovh|upcloud|digitalocean node-group list`) now renders:

```
default               type=4xCPU-8GB  count=3  joined=2  labels=1  taints=0  zones=fi-hel2
  no Kubernetes node is registered for: so-upcloud-infrastructure-cbd6f750-default-2
```

- `joined=N` appears only when it differs from `count`. A healthy group, a group the platform could not judge (empty node mirror, stopped cluster, AWS) and an older platform print the unchanged line.
- The second line names every counted worker with no node. A worker still creating lowers `joined` without being named, so a normal scale-up does not flag itself.
- `-o json` carries `joined_count` and `unregistered_workers` as served.

The remedy is not on the line on purpose: the phantom is not always the highest ordinal, and a scale-down removes the highest-numbered workers first, so it would take a real node.

## Tests

`cmd/cluster_node_group_list_test.go`: the mismatch line and the naming line for the customer's shape; the unchanged line for a healthy group and for an absent witness. `go build ./...` and the node-group list suites pass locally.


CodeStaple:
On the one finding: `cmd/cluster_node_group.go` already imports `strings` (the pre-existing `nodeGroupZoneSuffix` uses `strings.Join`), and the file builds and its tests pass in CI on this PR. No change.

PR #309, opened 2026-09-14
feat(bucket): ankra bucket create|list|get|delete (ankra-93ve0.5)

Myothas:
## What & why

A bucket for anything that isn't a backup vault (a container registry, application storage, a custody bucket) had to be made by hand in a provider console. The platform now manages object storage buckets as a resource (`/api/v1/org/object-storage-buckets`, ankraio/cluster#3046). This adds the CLI twin:

- **`ankra bucket create <name>`**
  - Creates a bucket on a Hetzner, Scaleway, DigitalOcean or UpCloud credential.
  - Defaults the credential to the only one Ankra can use (it refuses to guess between several) and the region to the provider's usual one.
  - `--wait` blocks until the bucket is ready or failed. Giving up waiting is reported as "still running", not as a failure.
  - Hetzner keys:
    - With no key flags, no keys are sent, so the platform uses the pair stored on the Hetzner credential (ankraio/cluster#3041).
    - `--access-key-id` prompts for the secret hidden, keeping it out of shell history.
    - Half a pair is refused before anything is sent.
- **`ankra bucket delete`**
  - Leaves the bucket and its objects on the provider by default.
  - `--destroy-provider-resources` names the provider-side bucket and provider in its prompt, and a declined prompt sends nothing (exit 4).
- **`ankra bucket list` / `get`** take a bucket name or id, support `-o json|yaml`, and never print key material. A bucket in `error` shows the excerpt and the recovery command.

Depends on ankraio/cluster#3046 being deployed. Until then the verbs answer 404 from the API.

Bead: ankra-93ve0.5

## Test plan

- `internal/client`: request shape for all four calls, including that absent keys and bucket are omitted (not empty strings) and that only `destroy=true` adds the query.
- `cmd`:
  - Name/uuid resolution, including ambiguity.
  - Create without keys on Hetzner, keys refused for UpCloud, half a pair refused, several fitting credentials refused.
  - Default delete leaves the bucket; the destroying prompt names the bucket and honours a "no".
  - `get` on an error bucket shows the excerpt and the way out.
- Pre-commit gate: `go test ./...` green, `golangci-lint` 0 issues.

## Docs checklist

- [x] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); ensure `Short`/`Long`/`Example` on new commands are written for docs readers


Myothas:
Addressed the Ankra AI review in 1a32e76:
- **Forced teardown of a provisioning bucket:** the platform enforces it (ankraio/cluster#3046: `objectstoragebuckets.Delete` sets `destroyProviderResources` for a bucket in `provisioning`, pinned by its parity suite). The command now reads the bucket before deleting, so the prompt says the half-made bucket is removed and the result says it is being destroyed, instead of promising the bucket stays.
- **--secret-access-key:** kept for automation. The flag help and the long help now say a flag value lands in shell history and the process list, and point at the hidden prompt you get by passing only --access-key-id.
- **Poll terminal states:** --wait now treats only `ready` and `error` as verdicts and keeps polling on anything else.

On CI: `TestClusterRoutesAreRegistered` fails because the new `/api/v1/org/object-storage-buckets` paths are only on ankraio/cluster#3046, not yet on cluster main's route census. An allowlist entry would go stale the moment that PR merges, so the order is: merge #3046, then re-run this PR's CI.

PR #315, opened 2026-09-14
fix(cluster): keep from_file through addons/manifests upgrade and encrypt (ankra-syg75)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-863/1185-cli-gaps-run-sha-defaults-to-local-head-validate-cannot-check-a
Bead: ankra-syg75

## What

Four CLI writers cleared the exported `from_file` before the partial-stack PATCH:

| Command | Site (origin/master@a686e59) |
|---|---|
| `cluster addons upgrade --set* / --values-from-file` | `cmd/cluster_partial.go:587` `applyAddonMutations` (since #71, ankra-u8ho) |
| `cluster manifests upgrade` (every mutation) | `cmd/cluster_manifests.go:331` |
| `cluster encrypt manifest` (cluster mode) | `cmd/cluster_encrypt.go:1215` |
| `cluster encrypt addon` (cluster mode) | `cmd/cluster_encrypt.go:1290`, rebuilt the configuration block without it |

They now carry it through. The addon dry-run notice changes from "values are sent inline, replacing the configuration.from_file reference" to naming the file the values land in.

## Why this is correct (code-read, cluster origin/main@3f23aa1c9)

On the PATCH lane `from_file` is a storage pointer, not a source:

- `go/internal/importedapi/specparse.go` `parseAddonConfiguration` / manifest parse: a plain optional string, no exclusivity with `values_base64` / `manifest_base64`.
- `enginekit/clusterengine/apistack.go:604-646` `apiAddonConfigurationDefinition` and `:661-693` `apiManifestDefinition`: `from_file` becomes `iac_from_file_path`; when absent it defaults to `stacks/<stack>/add-ons/<name>/values.yaml` / `stacks/<stack>/manifests/<name>.yaml`. `updateAPIAddonConfiguration` (`:1248`) preserves values and encrypted paths from the stored row, not the pointer.
- The GitOps push (`CalculateFullIaCFileSet`, `iac.go:846-887`) writes content to `resolveDumpFilePath(from_file, default)` and `cleanIaC` (`:975-983`) always strips `values_base64` / `manifest_base64` from the cluster document.

So a cleared pointer moved a hand-placed file to the default path and left the old copy stale. The reason #71 cleared it ("a stale pointer would contradict the inline values") holds on the IaC *apply* lane, where `from_file` is read; it does not hold on this lane.

Reported on PLA-863 (addon arm, `addons upgrade kyverno --set ... --dry-run`); the manifest arm was found on PLA-848.

## Tests

`cmd/from_file_pointer_test.go`: addons upgrade `--set`, its dry-run notice, manifests upgrade `--set`, encrypt manifest and encrypt addon in cluster mode. The fixture uses non-default pointer paths (a dropped pointer would otherwise still store a path, just the default one) and asserts the serialized PATCH body, since `from_file` is `omitempty`. All five fail on origin/master and pass here. `TestApplyAddonMutations_CarriesConfigurationBlock` now pins the pointer kept through a values replacement.

Gates run locally: `go build`, `go vet ./cmd/`, `go test -count=1 ./...` (all green), `gofmt -l` clean on touched files. `golangci-lint` is not installed on this machine; it runs in CI.

## Merge

Scary-list, a human merges: it changes where content lands in customer GitOps repositories and touches the `cluster encrypt` (secrets sealing) writers. The crypto path itself is unchanged.


Myothas:
**🤖 tech-agent** — answering the Ankra AI review of `c5ce0a59e171`

Both findings are about the same hunk, `runEncryptAddonCluster` in `cmd/cluster_encrypt.go`. The review flagged both as unverifiable from the hunk alone ("I cannot see the full struct literal"). Reading the whole function on this branch, neither holds — no code change on either.

### [high] Possible nil-pointer dereference on `mutatedAddon.Configuration` (`cmd/cluster_encrypt.go:1296`) — not valid

The literal the review could not see is three lines above the guard, and it is an unconditional assignment, not a field inside a larger literal:

```go
1290	mutatedAddon := *addon
1291	mutatedAddon.Configuration = &client.AddonConfigurationSpec{
1292		ValuesBase64:   base64.StdEncoding.EncodeToString([]byte(encryptedYAML)),
1293		EncryptedPaths: newPaths,
1294	}
1295	if addon.Configuration != nil {
1296		mutatedAddon.Configuration.FromFile = addon.Configuration.FromFile
1297	}
```

`mutatedAddon.Configuration` is non-nil on every path that reaches line 1296, whatever `addon.Configuration` is. The guard on line 1295 is not protecting the write target — it is protecting the *read* of `addon.Configuration.FromFile` on the right-hand side, which is the pointer that can legitimately be nil (an addon with no configuration block). So the asymmetry the review noticed is deliberate and is the correct way round.

### [low] Encrypt addon path may drop other `Configuration` fields (`cmd/cluster_encrypt.go:1295`) — not valid

`AddonConfigurationSpec` has exactly three fields (`internal/client/iac.go:30-34`):

```go
type AddonConfigurationSpec struct {
	ValuesBase64   string   `json:"values_base64,omitempty"   yaml:"values_base64,omitempty"`
	EncryptedPaths []string `json:"encrypted_paths,omitempty" yaml:"encrypted_paths,omitempty"`
	FromFile       string   `json:"from_file,omitempty"       yaml:"from_file,omitempty"`
}
```

The rebuild sets all three — `ValuesBase64` and `EncryptedPaths` in the literal, `FromFile` on line 1296 — so there is no fourth field left to drop. `copyAddonConfiguration` would be equivalent here, but it is not interchangeable: it copies `ValuesBase64` from the *original*, and this path must overwrite it with the freshly sealed document. Rebuilding states that intent, so I left it as is.

One caveat I will state rather than overclaim: this is only true for today's three-field struct. If `AddonConfigurationSpec` grows a fourth field, this hunk silently drops it, and the tests would **not** catch that — `TestRunEncryptAddon_ClusterModeKeepsValuesFromFile` (`cmd/from_file_pointer_test.go:141`) asserts that `from_file` reaches the PATCH wire, not that the whole block round-trips. Worth a reviewer's eye on whether that is a risk worth pre-empting with `copyAddonConfiguration` + an explicit `ValuesBase64` overwrite.

**Verification basis:** code-read at `c5ce0a59e171` (`cmd/cluster_encrypt.go:1290-1297`, `internal/client/iac.go:30-34`), plus the test suite run reported in the PR body. The `[high]` finding being answered rather than fixed is one of the reasons this PR stops for a human merge.

Myothas:

**PR:** https://github.com/ankraio/ankra-cli/pull/315 — stops four CLI writers (`addons upgrade`, `manifests upgrade`, `encrypt manifest`, `encrypt addon`) from clearing the exported `from_file` pointer before the partial-stack PATCH, so hand-placed GitOps files stay where the customer put them instead of being re-created at the default path.

**State:** `ankra-cli#315  OPEN  ready-for-review since 2026-09-14T22:33:42Z  ci=SUCCESS  review=none  ai-review=NEUTRAL  merge=CLEAN  (read 2026-09-14T22:34:22Z)`

**Why I did not merge it:** scary list, on two counts — (1) it changes **where content lands in customer GitOps repositories**, and (2) two of the four fixed writers are the `cluster encrypt` (secrets-sealing) PATCH builders. Separately, the Ankra AI review's `[high]` finding was answered rather than fixed, which is its own hold. Everything else on the gate passes: CI green, merge CLEAN, 179+/14- across 6 files, review footer covers the current head `c5ce0a59e171`.

1. **The core claim, which is a code-read and not empirical.** I argue that on the partial-stack PATCH lane `from_file` is a *storage pointer* (where the platform writes the file) and not a *source* the API reads — evidence in the PR body from cluster `origin/main@3f23aa1c9` (`apistack.go:604-693` mapping it to `iac_from_file_path` with a default, `iac.go:846-887` writing content to `resolveDumpFilePath`). Nobody has run a real PATCH against a live stack to confirm it. If that reading is wrong, this PR makes the CLI send a pointer the API will treat as an input.
2. **That reverting #71's behaviour is safe.** #71 (ankra-u8ho) cleared the pointer deliberately, reasoning that "a stale pointer would contradict the inline values". I believe that reasoning holds on the IaC *apply* lane and not on this one — but it is a deliberate prior decision being undone, so it deserves a second opinion.
3. **The two `cluster encrypt` hunks specifically** (`cmd/cluster_encrypt.go:1211-1215`, `:1290-1297`). The crypto itself is untouched — same `EncryptYAML`, same `verifyKeyEncrypted` loop, same `encrypted_paths`. What changes is that the sealed document's destination path now travels with it. Worth confirming that carrying `from_file` cannot land a sealed blob anywhere other than the path the customer already had.
4. **The `[high]` AI-review reply**, in the comment above: `mutatedAddon.Configuration` is assigned non-nil at `cluster_encrypt.go:1291`, three lines before the guarded write at `:1296`, so the flagged nil-deref cannot happen and the guard is protecting the read side instead. Please sanity-check that reading rather than take mine.

I have left it ready-for-review and will not merge it. Bead `ankra-syg75`; originating ticket PLA-863 (addon arm), with the manifest arm found on PLA-848.

PR #316, opened 2026-09-14
fix(pipeline): run never takes the commit from another repository's checkout (ankra-l1owk)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-863/1185-cli-gaps-run-sha-defaults-to-local-head-validate-cannot-check-a
Bead: ankra-l1owk

## Problem

`cmd/pipeline_run.go` `runPipelineDispatch` (since #264, in every tag from v0.15.0), with no `--sha`:

- **no `--ref`:** read the working directory's HEAD and branch, whatever `--application`/`--repository` named. `ankra pipeline run --application smartinsight` typed in a checkout of `commerce` dispatched smartinsight with commerce's commit and branch.
- **`--ref` given:** refused with `--sha is required`.

## Server facts this relies on (cluster origin/main@3f23aa1c9)

`go/internal/usecase/pipelines/provenance.go` `verifyRunProvenance` (cluster#2612, deployed 2026-09-05 00:06Z) runs before anything is written:

- empty `head_sha` → Ankra resolves the tip of `ref`, or of the repository's default branch, on the repository's host and runs that;
- a `head_sha` that is neither that tip nor an ancestor of it → 422 `ErrHeadNotOnRepositoryRef`.

Empirical: `TestVerifyRunProvenanceTrustsARefResolvedTipWithNoHeadSHAGiven`, `...FallsBackToTheRepositorysDefaultBranchWhenNoRefIsNamed` and `...RefusesACommitTheProviderNeverPlacedOnTheRef` pass at that ref. So the foreign-HEAD dispatch was refused server-side rather than run; the user got a confusing 422, and the CLI's `--sha` requirement was stale.

## Change

- HEAD is read only when there is no `--ref` **and** the checkout is the selected application's repository (`app_repo_owner`/`app_repo_name` match origin). The listing walk is extracted from `pipelineSelectorFromWorkingDirectory` into `checkoutApplications`, shared by both.
- Otherwise, including `--repository <id>` (no id to owner/name lookup exists) and an unreadable or partly read listing, the dispatch sends no sha and stderr says `Running at the tip of <ref | the repository's default branch>, resolved by Ankra at dispatch`.
- Long help and `--sha` flag help describe both sources.

**Contract change for the reviewer:** the old comment and `TestPipelineRunRequiresSHA` said "resolving a ref belongs to the trigger lane; a dispatch never runs whatever the platform stored last". The second half still holds: the tip is a live read from the host at dispatch, not a stored commit. The first half predates the server doing that read itself. `--ref release-2.0` from a checkout on main still never pairs main's commit with that ref; it now sends the ref alone.

Behaviour change for `--repository <id>` inside the matching checkout: it now runs the default branch tip, not local HEAD. Pass `--sha` or use `--application`.

## Tests

`cmd/pipeline_test.go`: the PLA-863 foreign checkout, a ref with no sha outside a checkout, the fail-closed arms (unreadable listing, repository id), a named ref inside a checkout, and the existing HEAD-read case now bound through the listing. The four behaviour tests fail on origin/master and pass here.

Gates run locally: `go build`, `go vet ./cmd/`, `go test -count=1 ./...` (all green), `gofmt -l cmd` clean. `golangci-lint` is not installed on this machine; it runs in CI.

## Merge

Not scary-list: CLI-only, and the provenance boundary it relies on is server-side and unchanged. Auto-merge once CI and the AI review are green, unless the review disputes the contract change above.

Docs follow-up after release: ankra-docs pages from docs#339 that say `run` reads HEAD.


Myothas:
**🤖 tech-agent** — reply to the Ankra AI review of 443e7848898d

### [low] Applications listing is walked twice per dispatch (`cmd/pipeline_run.go:109`)

Correct, and thank you — it is real. Answering rather than fixing it here, with a follow-up
bead filed (`ankra-x4qit`).

**Confirming the read you could not make from the diff:** the first walk's results are *not*
still in scope. `resolvePipelineSelector` (`cmd/pipeline.go:93`) returns only a
`client.PipelineSelector`; `pipelineSelectorFromWorkingDirectory` discards the ids it matched
(`cmd/pipeline.go:163` returns `matchedIDs[0]` alone). So `checkoutBuildsSelectedPipeline`
genuinely re-lists.

**Size of the cost:** `checkoutApplications` stops as soon as
`listing.Pagination.TotalPages <= page` (`cmd/pipeline.go:205`) at
`maxApplicationLookupPageSize = 100`, so for an organisation under 100 applications each
"walk" is a single GET — two instead of one, on a command that already makes a dispatch POST
and, with `--wait`, a poll loop. At 500 applications it is 10 GETs instead of 5.

**Why not in this PR** — each way of removing it costs more than it saves:

1. **Reading the flags to short-circuit is unsafe.** `ankra application pipeline run <id>`
   reaches the same `runPipelineDispatch` (`cmd/application_pipeline.go:49`) on a command
   that registers only `registerPipelineRunDispatchFlags` and
   `registerStructuredOutputFlags` — never `registerPipelineSelectorFlags`. So
   `GetString("application")` there returns `""` for a selector that came from a positional
   argument, "no flags given" would wrongly read as "inferred from the checkout", and the
   PLA-863 foreign-HEAD dispatch this PR exists to close would be back on that path.
2. **Threading provenance out of `resolvePipelineSelector`** touches its 16 call sites across
   6 files (`pipeline_definition.go`, `pipeline_findings.go`, `pipeline_logs.go`,
   `pipeline_artifacts.go`, `pipeline_run.go`, `pipeline_schedules.go`) for a low-severity
   read, and would take this PR well past the size at which it is still one reviewable fix.
3. **A package-level memo** puts cross-test global state into a package whose tests share a
   process and swap `apiClient` per case.

And the walk cannot simply be deleted: `checkoutBuildsSelectedPipeline` must also answer for
an explicit `--application`, which never walked at all.

`ankra-x4qit` records the shape worth doing instead — let `runPipelineDispatch` receive the
checkout provenance from its two callers rather than re-deriving it, so the unsafe
flag-reading shortcut is never needed.

Verification basis: code-read at `443e7848898d` for the call graph and the page-cap
arithmetic; the behaviour this PR changes is covered empirically by the four tests in
`cmd/pipeline_test.go` that fail on `origin/master`.

PR #318, opened 2026-09-14
fix(migrate): database volumes mount below the data directory, and --stack names the deployed stack (ankra-q2500, ankra-lkx9h)

CodeStaple:
Two bugs found today filming a real compose-to-Kubernetes migration onto a Hetzner cluster.

## ankra-q2500 (P1): the database PVC was mounted at the data-directory root

The docker module turned `db-data:/var/lib/postgresql/data` into a claim mounted at exactly `/var/lib/postgresql/data`. Every ext4 block volume (Hetzner, UpCloud, DigitalOcean...) carries `lost+found` at its root, so postgres's initdb refused - `directory "/var/lib/postgresql/data" exists but is not empty` - the pod CrashLoopBackOff'd, and `ankra migrate up` waited out its entire `--timeout`. `/var/lib/mysql` has the same hazard (mysql tolerates `lost+found`, mariadb warns).

For the workloads the module already recognises as databases (the same image-name detection the export uses: postgres/postgis/pgvector/timescaledb, mysql/mariadb/percona) the claim is now mounted one level down:

```yaml
volumeMounts:
  - name: db-data
    mountPath: /var/lib/postgresql/data
    subPath: data
volumes:
  - name: db-data
    persistentVolumeClaim:
      claimName: db-data
```

`subPath` over injecting `PGDATA` because it is engine-agnostic: the path inside the container stays where the image expects it and the same shape works for postgres, mysql and mariadb, where `PGDATA` would move postgres alone and leave every other engine to its own entrypoint flags. The reasoning is a doc comment on `databaseMountSubPath`.

- Non-database workloads are untouched, and so is a database volume that is not the data directory (a `/backups` mount stays as written).
- A source that already writes below the mount - `PGDATA` or `--datadir` set to a subdirectory, the hand-rolled workaround for this failure - keeps its mapping, so there is no double subdirectory.
- Every adjusted mount produces a warning naming the workload, the claim, the path and the subPath.

Operational note, also in the CHANGELOG: a stack converted before this change holds its data at the volume root, so re-converting moves the data directory down one level and the restore has to run again (`ankra migrate up` does that itself).

## ankra-lkx9h (P2): `--stack <name>` was ignored for the stack name

`ankra migrate up ./app --stack notes` printed `stack notes (new), namespace notes`, then `==> Deploying stack app ...`, and the stack landed as `app`: the conversion named the stack after the source project and `runMigrateUp` overwrote the plan's name with it. Only the namespace honoured the flag.

The requested name is now the single value end to end:

- `migrate.ConvertRequest` gained `StackName` (`stack_name` on the module wire protocol), and the docker module renders the stack under it, falling back to the project's own name when it is empty.
- `ankra migrate up` passes the plan's stack name into the conversion and no longer overwrites it, so the plan line, `cluster.yaml`, `==> Deploying stack ...`, the backup-vault import and the closing summary are one name. An explicit `--stack` goes through the same sanitiser as the directory-derived default, so the plan cannot promise a name the manifests do not use.
- `ankra migrate convert` gained the matching `--stack` flag. It is distinct from `--cluster-name`: `--cluster-name` is `metadata.name` (the ImportCluster), `--stack` is `spec.stacks[].name`.
- `performMigrateConvert` applies the requested name to a single-stack result, so a module written before `stack_name` existed still deploys under the name the user was shown. The contract is documented in `ankra migrate modules --help`, `examples/modules/README.md` and the example Procfile module.

## Also

`.gitignore` ignored `testdata/` repo-wide under the "Test binary" heading (which `*.test` already covers), so no Go test fixture could ever be committed - the golden manifest in this PR would have been silently untracked. Removed that line.

## How verified

- New unit tests: the subPath decision across postgres / mysql / mariadb / PGDATA-at-the-mount-root / PGDATA-below-the-mount / `--datadir`-below-the-mount / a non-data-directory database volume / a non-database workload, each asserting the mount path, the subPath and the warning.
- A golden for the rendered database Deployment (`internal/migrate/docker/testdata/database-deployment.yaml`, refreshed with `go test ./internal/migrate/docker -update-golden`), so the mount shape is reviewed as a diff.
- `TestRenderNamesTheStackWhenOneIsAsked`, `TestMigrateUpDeploysUnderTheRequestedStackName` (applied ImportCluster's stack name, the written `cluster.yaml`, the import's stack, the plan line, the deploy line and the summary) and `TestMigrateConvertNamesTheStackApartFromTheCluster`.
- `go test ./...` green and `golangci-lint run` clean (0 issues), both locally and again through the repo's pre-commit hook.

Beads: ankra-q2500, ankra-lkx9h


CodeStaple:
Thanks @ankraai - both findings addressed.

**[low] requested stack name silently ignored for multi-stack results** - fixed in 216e08b (pushed after the reviewed commit `fbd06ef`). `performMigrateConvert` now warns when the name cannot be honoured, naming the module, how many stacks it returned and the name that was asked for, with a unit test (`TestPerformMigrateConvertSaysWhenTheStackNameCouldNotBeApplied`) driving it through a stub two-stack module. Silently dropping the name is the shape of ankra-lkx9h itself, so it should not be possible to reintroduce quietly.

**[medium] subPath double-applies when the mount target is already the relocated data directory** - keeping the current behaviour; that case needs the subPath. If the source sets `PGDATA: /var/lib/postgresql/data/pgdata` **and** mounts `store:/var/lib/postgresql/data/pgdata`, then the volume root *is* the data directory again, so on a block CSI volume it carries `lost+found` again and initdb fails exactly as it did before this PR. The guard is deliberately "is the data directory the mount root?", not "did the source move the data directory?": only a data directory that sits *below* the mount (the `store:/var/lib/postgresql/data` + `PGDATA=.../data/pgdata` workaround) is already safe and is left alone. The suggested extra condition would reintroduce the crashloop for the mounted-onto-PGDATA layout. Pinned as a test case in f0d9b75 (`postgres with PGDATA mounted at the volume root`), next to the case that must *not* get a subPath.

On "breaking the restore path": nothing copies the source volume. `ankra migrate` carries databases as logical dumps (`export` writes `data/<workload>/<database>.dump`, the restore loads them into the running server through the backup vault), so the PVC is always a fresh, empty volume and the physical layout inside it is not part of the data path. The changelog note is about re-converting a stack that is already deployed: an existing pod's data sits at its volume root, and after a re-convert the engine looks one level down, which is why that note says to let the restore run again.

CodeStaple:
@ankraai on the [low] finding in the f0d9b75 review - **volume mounted at a parent of the data directory**: that shape is already safe, so there is nothing to warn about.

`store:/var/lib/postgresql` with `PGDATA=/var/lib/postgresql/data` (or the default, which is the same path) puts `lost+found` at `/var/lib/postgresql/lost+found`, *beside* the data directory. initdb inspects `PGDATA`, not the volume root, and `/var/lib/postgresql/data` does not exist yet: the entrypoint creates it (as root, before dropping to `postgres`) and initdb finds it empty. That is the same shape as the hand-rolled workaround this PR replaces - mount the volume one level above the data directory - and it is exactly what `subPath: data` synthesises for sources that did not do it themselves.

So the guard is doing the intended thing on that input. It reads "is the effective data directory the mount root?", and every layout where the answer is no - target a strict parent, `PGDATA` a strict child, a `/backups` volume that is not the data directory at all - is already safe and must be left untouched. `TestRenderMountsDatabaseDataDirectoriesBelowTheVolumeRoot` pins that branch with "postgres with PGDATA already below the mount" and "mariadb with --datadir already below the mount" (`wantSubPath: false`); the parent-mount layout is the identical predicate, one directory up. Adding a warning there would tell users that a working conversion is broken.

PR #322, opened 2026-09-15
feat pipeline: branches lists the latest run per branch, ankra-n8q38.2

CodeStaple:
## What & why

`ankra pipeline branches` lists one row per branch of a pipeline's repository,
with the newest run on each. The question a person opens a pipeline listing to
answer is usually "where does each branch stand", not "what were the last fifty
runs", and until now those were the same list with every branch mixed into it
(epic ankra-n8q38).

Bead: ankra-n8q38.2

Each row carries the latest run on the ref, the outcome of the run concluded
before it, how many runs the ref has ever had, and how many of those a newer
commit superseded - runs cancelled because something newer took their
concurrency group, which are history rather than failures to look at. The
repository's default branch leads the table and the rest follow by most recent
activity.

```
ankra pipeline branches [--application <name-or-id> | --repository <id>] [--all] [--limit] [--cursor] [-o json|yaml]
ankra pipeline list --latest-per-branch        # the same table under the listing command
ankra application pipeline list <id> --latest-per-branch
```

- `--application` defaults from the checkout's `origin` exactly as the rest of
  the family does - it goes through the shared `resolvePipelineSelector`.
- Branches with no run in the last fourteen days are hidden, with a one-line
  footer on stderr saying how many and that `--all` shows them. A page whose
  every row is stale says that, rather than "no pipeline branches found": those
  are different facts and only one of them is fixed by passing `--all`.
- `-o json|yaml` is the platform's page untouched, stale rows included. A script
  that asked for JSON filters it itself, and dropping rows from a
  machine-readable page would make `--all` a flag the script has to know about
  to be told the truth.
- `--latest-per-branch` cannot be combined with the run filters
  (`--status`, `--trigger`, `--branch`, `--head-sha`): they narrow runs, it lists
  refs. The combination is refused with exit 2 rather than silently ignoring the
  filter, and `--all` without it is refused the same way.

Columns: BRANCH / KIND / LATEST RUN # / STATUS / OUTCOME / PREVIOUS / RUNS /
SUPERSEDED / LAST ACTIVITY. STATUS is the latest run's lifecycle, OUTCOME its
verdict with the glyph the rest of the pipeline surface uses, and `-` for a run
that has not concluded - the STATUS column already says it is queued or running,
so repeating that in OUTCOME would cost the column its one job.

The `ankra-cicd` embedded skill now points at the command, so the agent guidance
and the CLI agree. (The monorepo's `ankra-skills` copy is not part of this PR;
the embedded copy is canonical in this repo and `make verify-skills` is a no-op
here.)

Requires the matching platform route, https://github.com/ankraio/cluster/pull/3092.
Against a platform without it the command answers the API's own 404.

## Test plan

- `cmd/pipeline_branches_test.go`: registration, every column present, the
  default-branch and pull-request labels, stale rows hidden with the footer,
  `--all` showing them and claiming nothing was hidden, `-o json` keeping every
  row and skipping the table, the empty page, the all-stale page, paging passed
  through, `list --latest-per-branch` reading branches and not runs, both
  refusals, and the sentinel error verbatim.
- One test drives the real `internal/client` against an `httptest` server via
  `useTestClient`, for the thing a mock cannot check: that the command asks
  `/api/v1/org/pipeline-repositories/{id}/pipeline-branches` with the paging in
  the query.
- The pre-commit gate ran in full on the commit: `go test ./...` green across
  every package (`ok ankra/cmd 13.902s`) and `golangci-lint run` at 0 issues.
  `go build ./...` green.

## Docs checklist

- [x] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); `Short`/`Long`/`Example` are written for docs readers


Myothas:
Re the review finding: fixed in 8763231. The next-page hint now prints for every page shape, including one whose every row was hidden as stale, with a test pinning it (`TestPipelineBranchesAllStalePageStillNamesTheNextPage`). The same push also makes the OUTCOME cell read `superseded` for a superseded latest run, matching `pipeline list` after #323.

PR #323, opened 2026-09-15
feat pipeline: a superseded run reads superseded and names the run that took its place, ankra-n8q38.1

CodeStaple:
## What this changes

A push that lands while the previous run is still going takes its concurrency group, and the older run is stopped. The CLI showed that run exactly as it showed one somebody had cancelled — `⊘ cancelled`, and nothing anywhere naming the run that had replaced it — so the person who pushed went looking for whoever had cancelled their build.

The platform now records the supersession as a fact (cluster PR below): `error_class = "superseded"` beside the cancelled outcome, plus `superseded_by_run_id` and `superseded_by_run_number` on the run response. This reads it.

- `ankra pipeline list` — the STATUS cell reads `superseded` instead of `cancelled` for a run a newer run replaced. The glyph and colour stay the ones every stopped run carries: it was stopped, and nothing about it needs attention.
- `ankra pipeline get` — prints `Superseded: by run #M` under the Status line, and says so without naming a run when the platform reports the class but no number (the newer run has since been removed by retention), never `#0`. A platform too old to report supersessions prints nothing, which is "this platform does not answer the question", not "this run was not superseded".
- `internal/client.PipelineRun` carries `superseded_by_run_id` and `superseded_by_run_number`, so `-o json` passes both through for scripts. `outcome` is still `cancelled`, so a script filtering on the outcome alone still finds these runs and reads `error_class` to tell them from a run somebody stopped — documented in `ankra pipeline get --help` and `ankra pipeline list --help`, and in the CHANGELOG's Unreleased section.

`renderPipelineState` is unchanged for steps; runs go through a new `renderPipelineRunState`, and both share one body so the glyph/colour table stays defined once.

Needs the platform change: https://github.com/ankraio/cluster/pull/3094

## Test evidence

The repo's own pre-commit gate ran green on this commit: `go test ./...` (all packages) and `golangci-lint run` (0 issues).

Five new tests in `cmd/pipeline_test.go`, all through the lane's established `runPipelineCommand` + `pipelineLaneMock` harness (the pipeline cmd tests drive a mock `APIClient` rather than an `httptest` server, so `useTestClient` does not apply here):

- `TestPipelineGetNamesTheRunThatSupersededIt`
- `TestPipelineGetSaysSupersededWithoutNamingARunTheServerDidNotReport`
- `TestPipelineGetOfACancelledRunNamesNoSupersession`
- `TestPipelineListShowsSupersededInTheStatusCell`
- `TestPipelineGetJSONCarriesTheSupersession`

Bead: ankra-n8q38.1

https://claude.ai/code/session_019swY1tpXkB4e2zkKtEiBdG


CodeStaple:
Both exist already and are unchanged by this PR, which is why they are not in the diff:

- `pipelineRunErrorClass(run client.PipelineRun) string` is defined at `cmd/pipeline_run.go:670` (added with the Class line for PLA-851/ankra-edt1b); it trims `run.ErrorClass` and answers `""` when the server recorded none.
- `client.PipelineRun.ErrorClass` is `*string` with tag `json:"error_class"` at `internal/client/pipelines.go:88` (`PipelineStep` carries the same field at :153).

Compile-level proof: the repo's pre-commit gate ran green on this commit (`go test ./...` across all packages plus `golangci-lint run`, 0 issues), and both `lint-test-build` CI runs for commit 5861207 are green.

Not merging from here: this PR reads fields that only exist once the platform change lands, so it should go in after https://github.com/ankraio/cluster/pull/3094 (which carries an alembic migration and is waiting on a human).

Myothas:
Re the review finding: `pipelineRunErrorClass` predates this PR (cmd/pipeline_run.go on master, added in #303) and is nil-safe: it returns "" for a nil `ErrorClass`, which is what lets an older platform print no Superseded line. Verified in the pre-release review before merging.

PR #326, opened 2026-09-15
fix: v0.17.0 pre-release review - migrate stack name, AWS bastion list and --force, backup block typos, restore/protect flag order (ankra-3vkh3)

Myothas:
## What & why

A pre-release regression review of everything merged since v0.16.1 (25 PRs) found six confirmed defects. All six are in features that shipped in v0.17.0-rc0 or after, so they go in before v0.17.0 rather than as follow-ups.

Bead: ankra-3vkh3

| # | Defect | Fix |
|---|--------|-----|
| #318 | `ankra migrate up` without `--stack` named the stack after the source **directory**, so a compose project with its own `name:` migrated on v0.16 re-deploys as a second stack (and a multi-stack module registered the restore against a stack that does not exist). | The default is the module's own name again (learned by one in-memory convert at plan time), the directory only as fallback. Namespace keeps the directory default, where earlier migrations put the workloads. `--stack` still names both. |
| #300 | `ankra cluster aws create\|preflight` without `--bastion-allowed-ips` sent `"bastion_allowed_ips": null`; the platform's `requiredStringList` refuses null as "not a list" (422). Only `[]` means its default. | An unset flag sends `[]`. |
| #301 | `ankra cluster deprovision --force` on an AWS cluster printed "warning: --force has no effect for this cluster type" while forwarding it. | AWS added to the honouring lanes and to the flag help. |
| #320 | `backup: {enable: true, vault: prod}` (any typo) parsed as a block with `enabled` unset and was sent as `enabled: false`, which the platform honours as a deliberate removal: **a typo unprotects the stack**. | Unknown keys in `backup`, `backup.retention`, `backup.selection` are refused with the file's vocabulary. |
| #320 | `addons settings set --backup-schedule=hourly` (the command's own example) on an add-on with no override created `{enabled:false, schedule:hourly}`: it stopped capturing the add-on. | A block created by a schedule/retention flag starts enabled; `--backup-enabled=false` still turns it off. |
| #325 | `restore-points restore` and `stacks protect` validated `-o` **after** the destructive write: `restore shop <id> --yes -o josn` removed the volumes, exited 2, and never printed the run id. | Both check `-o` before any request. |

Also from the review:
- **CHANGELOG** entries for #316 (`pipeline run` no longer requires `--sha`; three behaviour changes were unannounced) and #315 (`from_file` kept), both of which merged with no entry.
- The three embedded skills (`ankra-cicd`, `ankra-ship`, `ankra-applications`) still told an agent `--sha` is required outside a checkout; they now describe the tip-of-ref dispatch.

## Test plan

One new test per defect, each asserting the behaviour the review proved missing:
- `TestMigrateUpNamesTheStackAfterTheComposeProjectByDefault`
- `TestAwsCreateSendsAnEmptyBastionListWhenTheFlagIsOmitted` (create and preflight; asserts `"bastion_allowed_ips":[]` on the wire)
- `TestClusterDeprovision_ForceIsHonouredOnAws`
- `TestParseStackBackupRefusesAKeyItDoesNotRead` (four typo shapes)
- `TestSettingsSetScheduleAloneKeepsTheAddonCaptured`, `TestSettingsSetBackupEnabledFalseWinsOverTheDefault`
- `TestRestorePointsRestoreChecksTheOutputFormatBeforeDispatching`, `TestProtectChecksTheOutputFormatBeforeWriting`

Pre-commit gate ran in full: `go test ./...` green on every package, `golangci-lint run` 0 issues.

## Docs checklist

- [x] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); `Short`/`Long`/`Example` are written for docs readers
- [ ] Broader behaviour changes — docs updated in ankra-docs


Myothas:
Re the two findings:

- **Unknown keys in `backup:`** stay a refusal on purpose. The whole defect is that a mistyped key reaches the platform as a *narrower* policy (`enable: true` arrives as `enabled: false`, which unprotects the stack); a warning on stderr while still sending that narrowed block would not change what the platform does. The addon `configuration` block warns instead because its extra keys are ignored, not re-read as a different value. A new platform key does require a CLI update, which is the same trade the `retention`/`selection` typed fields already make.
- **Double `Convert`**: fair. The follow-up commit reuses the plan-time conversion for the write instead of converting twice, so the name announced in the plan is the name deployed by construction.

Myothas:
Re the second-pass findings:

- **Multi-stack modules**: the restore and the data import are keyed off `plan.Stack` (`performMigrateRestore` further down `runMigrateUp`), and that is exactly the seam this fixes. Before #318 it was `Stacks[0].Name`, a stack the deploy really creates; rc0 made it the directory name, which for a multi-stack module is a stack the cluster never receives. It is `Stacks[0].Name` again. Carrying data into more than one stack of a multi-stack conversion was never supported and is not claimed here; the changelog line describes the regression (a name that does not exist) rather than a new capability.
- **Strict keys**: the allow-list is the full vocabulary #320 defined from the platform's own `backup` policy shape (`enabled`, `vault`, `schedule`, `retention.{hourly,daily,weekly,monthly,yearly,minimum_count,minimum_age}`, `selection.{databases,confirm_exclude_databases,persistent_volume_claims}`), so a file the platform exports round-trips. A future platform key will need a CLI release, which is the same trade the typed `retention`/`selection` blocks already make, and cheaper than the alternative this fixes (a typo silently unprotecting a stack).

PR #330, opened 2026-09-15
feat(backups): ankra cluster backups status - what a whole cluster is backing up (ankra-0xsdd.46)

CodeStaple:
Bead: ankra-0xsdd.46

**Depends on [ankraio/cluster#3116](https://github.com/ankraio/cluster/pull/3116)**, which serves `GET /api/v1/org/clusters/imported/{cluster_id}/backups`. Merge that first.

## What this adds

`ankra cluster backups status` — the terminal twin of the portal's cluster Backups tab, and the one backup question the CLI could not answer.

`ankra cluster stacks restore-points list` answers per stack, and only about restore points: finding the stack nobody protected means already knowing to ask about it. `ankra backup vaults list` answers about vaults. Neither says whether a cluster's data is covered. Nothing in the CLI served the protection verdict, the schedule that produces it, or whether the backup data plane is even installed — so this is the gap rather than a second spelling of something that exists.

```
$ ankra cluster backups status --cluster production
Cluster production: 1 of 3 stacks with data are backed up, 1 could not be assessed, 1 had a backup fail in the last 24 hours.
Backup components: ready
Backup vaults: 1 of 2 ready.

╭──────────────────────┬────────────────────────────┬──────────────┬────────────┬───────────┬──────────────────────────────┬──────────────┬─────────────────────────╮
│ STACK                │ PROTECTION                 │ DATA         │ VAULT      │ SCHEDULE  │ LAST RESTORE POINT           │ NEXT RUN     │ LAST RUN                │
├──────────────────────┼────────────────────────────┼──────────────┼────────────┼───────────┼──────────────────────────────┼──────────────┼─────────────────────────┤
│ analytics            │ unprotected (no_policy)     │ 1 (0 db)     │ -          │ -         │ none                         │ not schedu…  │ none                    │
│ ankra-backup (Ankra) │ unprotected (no_data_asse…  │ 0 (0 db)     │ -          │ -         │ none                         │ not schedu…  │ none                    │
│ mystery              │ unknown (data_assets_unkn…  │ not reported │ -          │ -         │ none                         │ not schedu…  │ none                    │
│ shop                 │ protected                  │ 2 (1 db)     │ prod-vault │ 0 3 * * * │ complete 5.0 GiB, 1 day ago  │ in 14 hours  │ restore running, 2h ago │
╰──────────────────────┴────────────────────────────┴──────────────┴────────────┴───────────┴──────────────────────────────┴──────────────┴─────────────────────────╯
```

Flags: `--cluster` (or the selected one), `--protection` (repeatable), `--limit` / `--cursor` for the keyset page, `-o json|yaml`.

## Three-state answers, kept apart in the terminal too

- **The verdict has three values.** `unknown` is a stack Ankra could not assess — its inventory has not landed, or its backup settings could not be read — and it is never reported as unprotected. Both carry their reason in the cell, because "unprotected" and "unprotected because nothing in it stores data" are different facts.
- **`not reported` is not `0`.** A stack whose inventory has not landed reports zero assets, and zero assets is not "this stack holds no data".
- **An unreadable vault listing is not an organisation with no vaults.** `vaults_known: false` prints "could not be read" and deliberately withholds the "create one" nudge, so a blip on one small table does not tell an operator to set up something they already have.
- **The backup data plane has four states**, and `absent` and `degraded` say what they mean rather than printing a word: a cluster whose every stack names a ready vault and which has no backup stack is not a protected cluster.
- **`--protection partially` is refused** with exit 2 and the closed set, rather than sent and answered with an empty page that reads as "nothing matched".

The `ankra-backup` stack stays in the listing — so the counts are the platform's counts — and is marked `(Ankra)` so nobody has to wonder why the backup stack reports no protection.

## Files

- `internal/client/cluster_backups.go` — the typed client, bearer twin only, like `restore_points.go` and `backup_vaults.go`
- `cmd/cluster_backups.go` — the command
- `cmd/services.go`, `cmd/e2e_test.go` — the `APIClient` method and the `baseMock` stub the interface requires
- `cmd/cluster_backups_test.go`, `CHANGELOG.md`

## Gates

`make build`, `go vet ./...`, `go test ./...`, `golangci-lint run` (0 issues) — all clean.


CodeStaple:
@ankraai fixed in 2aced11.

**[low] `--limit` help promises a maximum the flag does not enforce** — you are right on both halves, and the negative case was the worse one: `--limit -3` was silently dropped by the client's `Limit > 0` guard, so the user got a page size they did not ask for and was never told. `clusterBackupsOptions` now refuses anything outside 1..100 with exit 2 naming the bound, the same shape `--protection` already had, and the ceiling is `client.MaximumClusterBackupsPageSize` so the help string and the check cannot drift. A test covers 0, -3 and 5000 and asserts none of them reaches the API.

While there: `describeBackupStackState` had the same silent-drop shape one level up — its default arm printed the raw value, and a bare word reads as a verdict the command stands behind. It now says the state is one this version of the CLI does not recognise, so a data plane in a state a newer backend introduced cannot be mistaken for `ready`. (The same finding was raised on the portal PR and fixed there too.)

On the build status you noted as unknown: `lint-test-build` passed on 9a27598, and locally `go build ./...`, `go vet ./...`, `go test ./cmd ./internal/client` and `golangci-lint run` (0 issues) were all clean before that push.

PR #331, opened 2026-09-15
docs(mesh): say that moving a cluster onto a fresh pod range is disabled (ankra-0xsdd.25.39)

Myothas:
Bead: ankra-0xsdd.25.39
Linear: https://linear.app/ankra/issue/PLA-867

cluster#3125 is merged (9ed535ff83), so this now describes live platform behaviour.

## Why

The day-2 pod range move crashes kube-controller-manager: its node IPAM allocator marks every Node's podCIDR as occupied at start and exits when one lies outside `--cluster-cidr`. Reproduced on kind, Kubernetes v1.34.0, kubeadm. cluster#3125 refuses the move in make-ready and in the update lane.

v0.17.0 shipped help text promising exactly that move, so the CLI currently tells people to run the thing that breaks their cluster.

## Change

Text only, no behaviour change - the platform refuses, and the CLI relays the reason as it already did.

- `mesh make-ready`: the help says the move is disabled and why, and that `--pod-cidr` still accepts the kubeadm default `10.244.0.0/16` as the step-back for a cluster an earlier move split. The `--pod-cidr` flag help says the same.
- The success line reads "Pod range recorded as ...: the cluster's nodes already run it, so no node moves" instead of announcing that every node leaves and re-registers.
- `mesh readiness`: an overlapping pod range is no longer listed among the things make-ready fixes in place.
- `mesh up`: `--renumber-pods` is described as having no effect until the move returns, and the plan line says the renumber would be refused.
- CHANGELOG entry under Unreleased.

The `--renumber-pods` planning path is left in place: it only triggers when the platform marks an overlapping pod range remediable, which it no longer does, and the redesign (ankra-0xsdd.25.36) is expected to restore it.


Myothas:
On the review of 1cc299b7ac5f. The branch is now rebased onto master (the CHANGELOG conflict with #330 was resolved by keeping both entries) as ddcd4c0.

- **[low] Plan output still suggests `--pod-cidr auto`.** Reworded. The plan line now reads `plan: <cluster> needs its pod range moved (<reason>), which the platform refuses while that move is disabled`, so it no longer spells out a command that would be refused. That branch only fires when the platform marks an overlapping pod range remediable, which it no longer does, so in practice `up` reports such a cluster blocked with the platform's own reason.
- **[low] The claim that the platform refuses the move is unverifiable from this diff.** The refusal lives in the platform, in cluster#3125 (merged as 9ed535ff83). `MakeMeshReady` returns a `MakeReadyRefusal` for any pod range that would move a running cluster, which the API answers as a 400 with that reason and this CLI already relays. The update lane also refuses to write a `--cluster-cidr` that excludes a live node's range. This PR only brings the help text in line with that.

PR #332, opened 2026-09-15
fix(apply): a cluster file that omits force/auto_remediate no longer clears them (ankra-mp2tr, PLA-834)

Myothas:
## What & why

The platform (cluster#2843) inherits a manifest's stored `force` / `auto_remediate` when the apply request omits the key, and takes the sent value, true or false, when it is present. ankra-cli#274 declared both as plain `bool` with no `omitempty`, so every `ankra cluster apply` of a file that never mentions the flags sent `"force": false, "auto_remediate": false`, which the platform honours as a deliberate clear. A `force: true` set through Git or the portal was wiped by the next routine apply (the PLA-834 shape). This has shipped in every release since v0.15.1.

Bead: ankra-mp2tr

- `client.Manifest.Force` / `.AutoRemediate` become `*bool` with `omitempty`.
- The apply parser returns nil for an absent or null key, a pointer for a boolean, and still refuses a non-boolean.
- No other construction or read site exists (the partial-PATCH lanes work on `ManifestSpec`, which has no such fields; nothing renders or exports them).

## Test plan

`TestBuildManifestApplyFlags` subtests: pass-through true (wire carries `"force":true`), **absent flags are omitted on the wire** (neither key present), **an explicit false is sent as false**, explicit null reads as absent, quoted/non-boolean still rejected. Pre-commit gate: `go test ./...` and `golangci-lint` green.


Myothas:
Re the findings: a repo-wide grep for `AutoRemediate` and for `.Force` on `client.Manifest` finds exactly one construction site (`buildManifest`) and no read site: the partial-PATCH lanes work on `ManifestSpec`, which has neither field, and nothing renders, diffs or exports them. `parseBoolField` has `buildManifest` as its only caller. Both confirmed by the build and the full test suite in the pre-commit gate.

PR #334, opened 2026-09-15
fix(registry): robots create prints the secret once; rotate/delete and related-repos remove prompt on stderr (ankra-1dp4t)

Myothas:
## What & why

`ankra registry robots create` printed the secret and then the platform's `docker_login` line, which embeds the same secret as `-p '<secret>'`: twice on screen, and pasting the line puts it into shell history and `ps`, the hazard `bucket create` warns about. The "shows the secret once" test only asserted `Contains`.

Bead: ankra-1dp4t

- The human output prints the secret exactly once, then `docker login <host> -u '<robot>' --password-stdin` built from the robot's own `Host`/`RobotName` fields (never the secret-bearing platform line; nothing is printed when either field is blank). `-o json|yaml` keep `secret` and `docker_login` untouched.
- `robots rotate`, `robots delete` and `org ai-review related-repos remove` now write their confirmation prompt to stderr, so `-o json` stdout is JSON only; each gains `-y` (the shorthand `cluster delete`, `alerts delete` and others already use).

## Test plan

`TestRegistryRobotsCreateShowsTheSecretOnce` now counts occurrences (exactly 1) and asserts the login line carries no secret; `TestRegistryRobotLoginCommandNeverCarriesTheSecret`; `TestRegistryRobotsRotateJSONKeepsThePromptOffStdout`, `TestRegistryRobotsDeleteJSONKeepsThePromptOffStdout` (stdout unmarshals as JSON, prompt on stderr); `TestRelatedReposRemovePromptStaysOffStdout`. Pre-commit gate green.


Myothas:
Re the findings:

- **Host unquoted**: fixed in 5b30366, the host is single-quoted the same way as the login (`shellSingleQuote`), and the test pins the full line.
- **RobotName semantics**: it is the full login. `internal/client/registry_robots.go` decodes `robot_name` separately from `name`, and the client test asserts `RobotName == "robot$org-abc+user-jenkins"` for a robot created as `jenkins`.
- **Test buffers**: `executeCommand` calls `rootCmd.SetOut/SetErr` with a fresh buffer on every call, so a leaked buffer from the split helper cannot reach another test.

PR #342, opened 2026-09-15
fix(skills): a partial uninstall keeps the recorded install options (ankra-d3fpg follow-up)

Myothas:
## What & why

Follow-up to #340, addressing its Ankra AI review (changes requested):

- **[high]** `uninstallForTarget` forgot the recorded install options on every personal-scope uninstall, but it is also called with `full=false` to remove only the named skills while leaving the install (rule block, workflows, hook) in place. After such a partial uninstall the next `ankra upgrade` refreshed that client with the defaults, re-adding what the person declined, which is exactly the defect #340 fixed. The forget is now gated on a full uninstall.
- **[low]** A forced `--skills` refresh after a detection failure printed "Refreshing the Ankra agent skills for  ..." with no names and ran nothing. With no clients it now says nothing was found and names the by-hand command.

Bead: ankra-d3fpg

## Test plan

`TestSkillsInstallRecordsTheOptionsItUsed` now does a partial uninstall first and asserts the record survives, then the full one and asserts it is forgotten; new `TestRefreshInstalledSkillsWithNothingDetectedSaysSo`. Pre-commit gate green.


Myothas:
Re the finding: reworded in ef18df0. The message now names both possibilities (no assistant detected with an install for this user, or an unreadable record, which the warning above already reported) and no longer asserts that nothing is installed or suggests reinstalling over a valid install. The test pins that the phrasing never says 'were found'.

PR #343, opened 2026-09-15
fix(pipeline): name a retried step's earlier attempts and how to read their logs (ankra-mlf7a)

Myothas:
Linear: https://linear.app/ankra/issue/PLA-871/1193-pipeline-build-lane-buildkit-gets-500-pushing-to
Bead: ankra-mlf7a

## What went wrong

On PLA-871 (Smartoptics), commerce run #193 finished green, and both of its
build steps showed a first attempt concluded `infra_error` followed by a
successful automatic retry. The reporter asked what the first attempt was and
could not find out:

> `ankra pipeline logs --step build-commerce-backend` only returns the
> successful attempt.

That reading was right, and the failed attempt's log was not lost — it was
unnameable.

- A retried step is several rows under one key. The lost attempt is kept on
  the run as evidence (`enginekit/pipelinerun`'s `insertRetryAttempt`), and
  its log is uploaded like any other: the platform builder PUTs its transcript
  on **both** outcomes (`scheduler/go/cmd/platform-builder-agent/main.go`,
  `uploadBuildLog`), and an in-cluster step archives one `step_log` per
  attempt row (`scheduler/go/internal/maintloops/pipeline_artifacts.go`).
  Both grants are keyed on the attempt's own step row id.
- `resolvePipelineStep` (`cmd/pipeline_logs.go`) already answers a step **id**
  exactly — "that is how a lost attempt's own log stays readable" — and
  resolves a step **key** to the newest attempt. Correct, and silent.
- `printPipelineRunDetail` (`cmd/pipeline_run.go`) prints one table row per
  step row, with columns STEP / STAGE / KIND / STATUS / EXIT. So a retried
  step showed two rows that differed in nothing a reader could see, and
  neither carried the attempt number or the row id that `--step` accepts. The
  attempt's own `error_class` and `error_message` — already in the run detail
  payload — were never printed at all.

The only path to a lost attempt today is `ankra pipeline artifacts <run>`
(whose STEP column happens to be the step row uuid) plus `artifacts download`.

## What this changes

`ankra pipeline get <run>`:

- the step table gains an `ATTEMPT` column, so two rows under one key are
  distinguishable;
- below it, a run with a retried step prints an **Earlier attempts** block:
  one entry per superseded row with its outcome, its own error class and
  message, and the exact `ankra pipeline logs <run> --step <attempt id>`
  command that reads its log. Nothing is printed for a run with no retried
  step, which is almost every run.

`ankra pipeline logs --step <key>`:

- still resolves to the newest attempt — that is the attempt that did the work
  — and now says so on stderr when earlier attempts exist, listing the command
  for each. stdout stays the log and nothing else, so redirecting it to a file
  is unchanged.
- `--help` and the `--step` flag text say that an id names one attempt.

No server change; no behaviour change for a run whose steps were not retried.

## Tests

- `TestPipelineRunDetailNamesASupersededAttempt` — ATTEMPT column, the block,
  the superseded attempt's class and message, the log command naming attempt
  1's row id, and that the newest attempt is not listed (its key already
  answers for it).
- `TestPipelineRunDetailSaysNothingAboutAttemptsWhenNothingWasRetried`.
- `TestPipelineLogsByKeyNamesTheEarlierAttempts` — stdout is the resolved
  attempt's log, stderr names the earlier one and its command.
- `TestPipelineLogsSaysNothingAboutAttemptsForAStepThatWasNotRetried`.
- `TestPipelineLogsByAttemptIDReadsThatAttemptsOwnLog` — the path the note
  points at downloads the lost attempt's own `step_log`.

Gates run here: `go build`, `go vet`, `go test -count=1 ./cmd/...`,
`gofmt -l`. `golangci-lint` is not installed on this machine; the full linter
runs in CI.

Refs read: `ankra-cli` origin/master@27a22ea, `cluster` origin/main@57eed9f35.


Myothas:
**🤖 tech-agent** — both review findings fixed in da64641

**[medium] Suggested log command omits the --application selector** — real, fixed.

You were right to doubt the hint, and the reason is slightly different from the one you
guessed: `--application` is not *mandatory*. `resolvePipelineTarget`
(`cmd/pipeline.go:120-169`) accepts `--application`, or `--repository`, or neither — in
which case it infers the selector from the working directory's git origin, and only if
that infers nothing does it fail with `one of --application or --repository is required`.

So the printed command was pasteable by exactly one reader: someone standing in the
matching checkout. Every reader who had named their application explicitly — which is
every caller the note was written for, and the reporter on PLA-871 — would have pasted it
and got a usage error. That is worse than printing no hint.

Both hints now carry the selector (`pipeline_run.go:712`, `pipeline_logs.go:658`), rendered
by a new `pipelineSelectorArguments` from the **resolved** `client.PipelineSelector`
rather than from the invocation's flags. That is deliberate: the flags are not the only way
a selector arrives. `ankra application pipeline logs <application-id> <run>` takes it as a
positional argument and registers no selector flags at all
(`cmd/application_pipeline.go:228-243`), and the inferred case has no flag by definition.
Reading the resolved selector answers for all three, and it fixes the inferred case the
right way round — the inference only works from inside the checkout, so a command printed
for that reader must spell the selector out to survive being pasted anywhere else.

An empty selector still prints no flag rather than an empty one
(`TestPipelineRunDetailPrintsNoSelectorItWasNotGiven`).

**[low] Superseded attempts listed in payload order, not attempt order** — real, fixed.

I could not find an ordering guarantee either, and this lane had already decided not to
rely on one: `newestPipelineStepAttempt` compares attempt numbers rather than take the
listing's order, on the stated grounds that "which row answers for the step is not a detail
worth inheriting from a server-side ORDER BY". Printing the chronology block in payload
order was the same trust, in the one place a reader is reading *for* sequence.

Both listings now sort: the run detail's by `(StepKey, Attempt)` (`pipeline_run.go:703`),
the `logs` stderr note's by `Attempt` (`pipeline_logs.go:651`), both stable.

`TestPipelineRunDetailOrdersEarlierAttemptsByAttempt` hands three attempts over
newest-first. I checked it is not a vacuous test: with the sort removed it fails, printing
`build attempt 2:` above `build attempt 1:`.

**Gates run locally on da64641:** `go build ./...`, `go vet ./...`,
`go test -count=1 ./cmd/...` (ok, 46s), `gofmt -l cmd/` (clean). `golangci-lint` is not
installed on this machine; the full linter runs in CI.

Myothas:

**PR:** https://github.com/ankraio/ankra-cli/pull/343 — `ankra pipeline get` and
`pipeline logs` now name a retried step's earlier attempts and print the exact command that
reads each one's log (PLA-871, bead `ankra-mlf7a`).

**State:** `ankra-cli#343  OPEN  ready-for-review since 2026-09-15T23:39:25Z  ci=PENDING
review=none  ai-review=NEUTRAL  merge=UNSTABLE  (read 2026-09-15T23:39:52Z)`

**Why I did not merge it:** size. My merge gate stops at 400 changed lines; after the
review fixes this is **409 additions / 16 deletions across 6 files (425 changed)**. CI was
still running when I wrote this and does not change that answer, so I am handing it over
now rather than after green. Nothing else in the gate fails: it is not on the scary list
(four `cmd/` files plus two test files — CLI output only, no schema, no deploy path, no
auth, nothing under `.github/`), it merges cleanly, no human has requested changes, and
both Ankra AI review findings are fixed in da64641 and answered above.

1. **The selector now printed inside the suggested command** (`pipelineSelectorArguments`,
   `cmd/pipeline.go`). The hint prints the *resolved* application/repository id, not the
   friendly name the caller typed. That is what makes it paste correctly for
   `ankra application pipeline logs <application-id> <run>` and for the
   working-directory-inferred case, but it does mean a reader who typed
   `--application commerce` is handed back a uuid. If you would rather echo their own
   wording when they used the flag, say so and I will split that out.

2. **That the "Earlier attempts" block is the output you want** — it is the part a customer
   reads on PLA-871. It prints one entry per superseded row with its outcome, its own error
   class and message, and the log command. `printPipelineSupersededAttempts` in
   `cmd/pipeline_run.go`.

3. The bulk of the size is tests and comment prose rather than logic — the whole change is
   print statements, two sorts, and one new 12-line helper. Worth a skim on that basis
   rather than on the line count.

Verification basis: **empirical** — `go build ./...`, `go vet ./...`,
`go test -count=1 ./cmd/...` and `gofmt -l cmd/` all run locally on da64641, and the new
ordering test confirmed failing with its sort removed. I have not run the binary against a
live run; the retry behaviour it renders is reproduced from fixtures.

PR #344, opened 2026-09-15
feat(stacks): ankra cluster stacks deploy-draft (ankra-0xsdd)

CodeStaple:
## The gap

Everything that builds a stack without running it leaves a **draft**: a clone, a stack-profile instantiation, `ankra cluster draft -f`. Deploying one was a **portal-only** move.

`ankra cluster stacks clone --deploy` is a *clone-time* flag. It cannot be replayed — a second clone onto the same target is refused because the stack name is taken — and with `--with-data` it only **plans** the deploy and waits for a restore. So the command-line path stopped one step short of a running stack.

That is what the third live verification pass hit on item 2, "the clone's data half lands on the target" (`docs/runbooks/backups-verification-2026-09-15-third-pass.md`, "What item 2 will hit next"): *"the draft deployed from the portal until a CLI verb exists."*

## The verb

```
ankra cluster stacks deploy-draft <stack name>
```

| | |
|---|---|
| `<stack name>` | positional, required — the stack that exists on the active cluster only as a draft |
| `--cluster` | inherited persistent flag; otherwise the selected cluster |
| `-o, --output` | `json` \| `yaml` via the shared structured-output flags |

Exit codes follow the CLI's scripting contract: `3` when the stack is not found (and the message names the stacks that *do* have drafts), `2` when the named stack has nothing to deploy, `1` when the platform refuses the write.

## No route was invented

There is **no** "deploy this draft id" route on the platform. What the portal's Deploy button does — and what the clone's own `--deploy` does in-process (`deployCloneDraft`) — is the ordinary create-stack write with the draft's id carried on the stack:

```
POST /api/v1/org/clusters/imported/{cluster_id}/stacks
{"spec": {"stacks": [{ ...the stack..., "draft_id": "<uuid>" }]}}
```

That path **is** the bearer-PAT twin of the browser route (`cliapi.go`, `importedPrefix+"/stacks"`), so the CLI addresses exactly the path the portal does. The server loads the draft's resources, sees the named stack is the draft bound to that id, and promotes it.

## Why the stack is passed as raw JSON

The create-stack write is **not partial**: `ProcessAPIStack(..., partial=false)` calls `removeDeletedAPIStackResources`, so members the request omits are **deleted** from the stack. The request therefore has to carry the draft's content *exactly*.

So the CLI never re-models it. The listing's stack object is kept as `map[string]json.RawMessage` and posted back unchanged except for one member:

- **`state`** is dropped. The listing renders `state: "draft"` for a draft-only stack and each of its members, and the spec parser's enum is `creating|updating|stopping|up|down` — posting the object unedited is a 422 on the first manifest.

Decoding into a struct instead would have dropped **`encrypted_paths`**, which the CLI's existing display structs (`StackManifest`, `StackAddonConfig`) do not carry. A stack deployed without its SOPS declaration seals nothing. Raw pass-through makes that class of loss impossible rather than merely unlikely, and it is pinned by a test that asserts `encrypted_paths`, `manifest_base64`, `values_base64`, `registry_url`, `parents`, `variables` and `resource_id` all survive verbatim.

## Refusals answered before the write

Two cases are refused here rather than by the platform, because the platform's wording was written for a different caller:

- a **clean deployed stack** has no draft to deploy;
- a **deployed stack holding a draft of edits** (`deployed_dirty`) is the *update* write, not this one. The create write refuses it with *"use update_cluster_stack instead of create_cluster_stack"* — naming an API function no CLI user has. This command points at the stack builder instead.

## Tests

Four client tests and six command tests, all new. `go build ./...`, `go test -count=1 ./...` and `golangci-lint run` (0 issues) are green; `-race` is CI's (no accepted Xcode licence locally).

## Companion

The clone-side defect found in the same pass — a clone carrying the source cluster's generated ingress hostname with no warning — is ankraio/cluster#3167.

Bead: ankra-0xsdd-clone-source-hostname


CodeStaple:
Thanks — all three are the right questions to ask about code the diff does not show. Two are answered with evidence, the third was a real gap and is fixed.

## [medium] CSRF — not required on this path, and a shipped sibling proves it

`/api/v1` is the bearer-PAT surface and has no CSRF requirement. `go/internal/cliapi/cliapi.go` says so in the package doc:

> Every route authenticates with `authenticate_user_token` (`Authorization: Bearer <PAT>` plus the optional `X-Ankra-Organisation-Id` override) **and has no CSRF requirement**.

The `postCSRFJSON` helper is also weaker evidence than it looks — it is a double-submit the CLI satisfies unilaterally (`csrf.go:56-61` generates the token *and* sets the matching cookie), so it is a formality on routes that want it, not a server-issued secret.

The decisive point is a sibling that already ships: `CloneStackToCluster` (`internal/client/stacks.go:331`) POSTs to `/api/v1/org/clusters/imported/{id}/stacks/clone` — the same prefix, one segment over — with a plain `Authorization: Bearer` and no CSRF header. `DeleteStack` does the same. Both were exercised against production in the third live verification pass. If this prefix required CSRF, `ankra cluster stacks clone` would already be broken.

## [medium] Other render-only members — the parser reads by name and ignores the rest

I checked every member you named against `go/internal/importedapi/specparse.go`. References in the parser:

| member | read by the write? |
|---|---|
| `is_draft_only`, `lifecycle`, `has_encrypted_values`, `version_history`, `jobs`, `job`, `draft`, `health`, `intent_requested_at` | **0 references — ignored** |
| `state` | read, and enum-validated to `creating\|updating\|stopping\|up\|down` — the one that must go |
| `resource_id`, `chart_icon`, `force_delete` | read — and **meant** to survive; `resource_id` is what binds each entry to the drafts existing resource on promotion |

Your `jobs` worry specifically: `parseAPIStack` never reads a stacks `jobs` (zero hits for `"jobs"` and no `.Jobs` assignment in the parser), so it cannot be honoured. And stripping the set explicitly, as you suggested, would have been the riskier move — `resource_id` is in that visual group and removing it would break the promotion.

## [low] Member arrays beyond the three — you were right, fixed

This one was a genuine gap: the fixtures stack-level `jobs` array was exactly the case the three-name list missed, and a member list added later would have 422d the same way the top-level `state` did.

Fixed by making the rule generic instead of enumerated — "strip `state` wherever it appears", applied to every array-of-objects member. It cannot remove anything the write needs, because `state` is the one member the write refuses, and members that are not arrays of objects (`variables`, `encrypted_paths`, scalars) never decode and pass through as the exact bytes the server sent. Two new tests pin both halves, including a `future_members` array standing in for a member list this client has never heard of.

CodeStaple:
Second pass — both new findings were real and are fixed in 9171949; the CSRF one is the same finding as last round and is answered below with the evidence again.

## [low] `is_draft_only` absent read as deployed-with-edits — fixed

You were right, and the failure mode is worse than a miss: the "no" branch is a **refusal**, so a genuinely draft-only stack whose listing omitted the boolean would have been refused with wording about a state it is not in ("already deployed and holds a draft of unsaved edits"). An absent observation was being read as a negative answer.

`IsDraftOnly()` now falls back to `lifecycle == "draft_only"` when the boolean is absent *or* fails to decode — the two are rendered from the same fact. Put in the client accessor rather than at the call site so both readers agree. Six-case table test covers present/absent/undecodable against both lifecycles.

## [low] Pagination truncation — fixed

Also real. `total_pages` decodes as `0` when the block is absent, so the walk stopped after page 1 and a draft on a later page was reported as **not found** rather than deployed — again a miss surfacing as a different state.

A short page now ends the walk whatever the pagination block says; `total_pages` is honoured only when present; and a page cap keeps a server that answers a full page forever from spinning. Two tests: one proves a full page is followed by the next when pagination is absent, the other that a short page still ends it in one request.

Worth noting the same truncation exists in the pre-existing `ListClusterStacks`, which I have deliberately **not** touched: that is a display path where a missing stack is visible to the person reading the table, not a lookup whose miss is reported as a different state. Happy to fix it separately if you would rather it were uniform.

## [medium] CSRF — unchanged answer, restated

`/api/v1` is the bearer-PAT surface. `go/internal/cliapi/cliapi.go` package doc: *"Every route authenticates with `authenticate_user_token` (`Authorization: Bearer <PAT>` …) **and has no CSRF requirement**."*

And `postCSRFJSON` is weaker evidence than it looks — `csrf.go:56-61` generates the token *and* sets the matching cookie, so it is a double-submit the CLI satisfies unilaterally, not a server-issued secret.

The decisive point remains a shipped sibling: `CloneStackToCluster` (`internal/client/stacks.go:331`) POSTs to `/api/v1/org/clusters/imported/{id}/stacks/clone` — same prefix, one segment over — with a plain `Authorization: Bearer` and no CSRF header, as does `DeleteStack`. Both were exercised against production in the third live verification pass. If this prefix enforced CSRF, `ankra cluster stacks clone` would already be broken.

CodeStaple:
Third pass. Two fixed in 59cf2f8 (above), two answered here.

## [medium] Pagination stops early if the server caps page_size — fixed, and you caught a regression I introduced

You were right and it was worse than a latent risk: my *previous* commit traded one truncation for another. Ending the walk on any page shorter than the size asked for is exactly wrong against a server that silently caps `page_size` — every page is then "short" and the walk stops after page 1, reporting a draft on a later page as not found, which is the failure the guard was added to prevent.

`total_pages` is now authoritative whenever the server sends it, and the page-length rule applies only when the block is absent. Three tests now cover the three shapes: absent pagination with a full page (walks on), absent pagination with a short page (stops in one request), and `total_pages: 3` with the server capping pages at 25 (walks all three and finds the draft on the last).

## [low] Deploy runs on context.Background() — fixed

Also right. It now uses `cmd.Context()`, as `cluster_addon.go` and `cluster_decrypt.go` do, so Ctrl-C cancels the request instead of leaving the writes outcome unknown.

## [low] State strip does not recurse into nested arrays — no change, and recursing would be wrong

Checked the listing render. Nested entries inside a member entry never carry `state`:

- `parents` entries render `{name, kind}` only (`apirender.go`, `parentsWireValue`) — no state member exists;
- `version_history` renders `WireArray()` — a literal empty array, on every member;
- `job` is an object, not an array, and its members are `{job_id, name, status, created_at, updated_at, superseded}` — `status`, not `state`. The writes parser never reads `job` at all.

More to the point, recursing at every depth would be *unsafe*, not merely unnecessary. An addons `settings` is a pass-through object (`SettingsRaw`), so a key literally named `state` inside it would be the users own value — and a blanket recursive strip would silently delete it. The current depth is where the writes enum actually bites, and stopping there is deliberate.

## [medium] CSRF — third time asked, same answer

`sendJSONContext` is in `internal/client/helpers.go:91-140` (unchanged by this PR, hence absent from the diff): it sets `Content-Type` and `Authorization: Bearer` and nothing else. That is correct for this surface — `go/internal/cliapi/cliapi.go` package doc says every `/api/v1` route *"authenticates with `authenticate_user_token` … and has no CSRF requirement"*, and `postCSRFJSON` is a double-submit the CLI satisfies unilaterally (`csrf.go:56-61` mints the token and sets the matching cookie), not a server-issued secret.

The shipped sibling remains the decisive evidence: `CloneStackToCluster` (`internal/client/stacks.go:331`) POSTs to `/api/v1/org/clusters/imported/{id}/stacks/clone` — same prefix, one segment over — with a plain bearer header and no CSRF, as does `DeleteStack`. Both ran against production in the third live verification pass. If this prefix enforced CSRF, `ankra cluster stacks clone` would already be broken.

PR #345, opened 2026-09-16
feat(backup): ankra backup vaults contents reads what a vault actually holds (ankra-0xsdd.72)

CodeStaple:
## Why

Every other backup command reports Ankra's own rows. This one lists the **bucket**, which is the only way to find out whether deleting a restore point actually removed anything.

It did not. A live verification on 2026-09-16 deleted a complete restore point, was told its objects were being swept, and found **all nine of them and 36 MB still there** — discoverable only by going around the product and listing the bucket by hand with the cluster's own vault credential. Four verification passes in a row had recorded "the sweep reported success" as a pass.

Needs the platform side: **ankraio/cluster#3173** (`GET /api/v1/org/backup-vaults/{vault_id}/contents`).

## The verb

```
ankra backup vaults contents <vault-name|vault-id>
    --prefix <p>           only read objects under this prefix
    --restore-point <id>   report one restore point (accepts an id prefix)
    --orphans-only         only the objects no restore point accounts for
    --objects              list every object key the listing read
    -o json|yaml           structured output
```

Per restore point it prints **two** prefixes, because they differ:

- **declared** — the prefix the restore point records, and that a delete sweeps
- **located** — where the backup data plane actually wrote

Volume data lives in a repository **shared** by every backup of a cluster and namespace, so its bytes are reported separately and never counted into a single restore point: no restore point owns them, and deleting that prefix would destroy other restore points' data.

`--orphans-only` shows the leak shape directly — a restore point Ankra deleted whose objects survived it, with the date the row went.

## Honesty properties

- **An unreadable vault fails** (non-zero exit) rather than printing an empty listing. An unreadable vault and an empty vault are different facts, and four passes in a row recorded the difference as a pass.
- **Orphan detection is withheld, loudly, whenever the read was partial.** The CLI keys off the platform's own `orphans_determined` field rather than inferring it from `complete`, so a narrowed read (`--prefix`, `--restore-point`) can never print "every object is accounted for" over a listing that never looked. Inferring it was a real bug in this PR, caught in self-review.
- A restore point with **nothing** found renders as `(nothing found)`, not as an empty cell that reads as "small".

## Tests

Mock-driven, against the shapes the fourth pass measured: both prefixes rendered, the shared repository kept separate and labelled unattributable, a surviving deleted restore point rendered as an orphan, "no orphans" vs "not determined" kept apart, an unreadable vault exiting non-zero, flags reaching the request, and a vault **name** resolving to its id.

The full `go test ./...` suite and `golangci-lint` both ran green via the pre-commit hook.

Bead: ankra-0xsdd.72


CodeStaple:
Both findings were right and both are fixed in 250d7d7.

**[medium] Partial-listing totals printed as definitive** — agreed, and this was the more serious of the two because it is the same conflation the command exists to remove, reintroduced one layer up. I had added an explicit `orphans_determined` field for the orphan section and then left the headline totals reading off nothing at all. The header now prints

```
PARTIAL:  this listing stopped short - the counts above are a FLOOR, not a total.
```

keyed off `Complete` itself, **not** off `Warnings` — exactly your point that a future partial cause carrying no warning would otherwise render as a full inventory. Covered by `TestBackupVaultsContentsMarksPartialTotalsWithoutRelyingOnWarnings`, which sets `Complete=false` with `Warnings=nil` precisely to pin that independence, and by a companion test asserting a complete listing is *not* marked partial.

**[low] Truncated restore point IDs may collide** — agreed. You flagged that you could not see the server-side prefix matching, so to close that half: the platform resolves `restore_point_id` as a prefix and **refuses an ambiguous one** with a typed `AmbiguousRestorePointError` (400) listing the candidates, rather than resolving it to the newest match — deliberately, because silently reporting on a restore point the caller did not name is worse than refusing in a verification tool. So a colliding id could not have mis-resolved. But your underlying point stands: a table that cannot distinguish two of its own rows is no good to read, and the help does invite copying the value. `restorePointIDWidth` now extends the width until every restore point in the listing is distinct, starting at 8. Covered by `TestRestorePointIDWidthExtendsUntilUnique` and `TestBackupVaultsContentsDisambiguatesCollidingRestorePointIDs`.

Full `go test ./...` and `golangci-lint` green via the pre-commit hook.

CodeStaple:
## CI is red for the right reason: this PR must merge *after* the platform side

`TestClusterRoutesAreRegistered` fails with:

```
1 /api/v1 path(s) the cluster router does not register:
  /api/v1/org/backup-vaults/%s/contents  (internal/client/backup_vault_contents.go)
```

That is correct and it is the gate doing its job. It fetches `routes.json` from **`ankraio/cluster` at ref `main`**, and the route this client calls lives on the unmerged branch behind ankraio/cluster#3173. So right now the CLI really would be calling a path production does not serve.

**I am not allowlisting it.** The allowlist is explicitly reserved (in the test's own doc comment) for paths the scan *cannot verify* — a `%s` holding a literal segment chosen at run time. A route the cluster genuinely does not register is the other kind, and the doc is unambiguous about it: it "comes off this list when the route exists on the cluster side, not before". Since the ratchet is shrink-only, an entry would also fail a second time the moment #3173 lands, as "allowlisted path now resolves; drop it". Silencing it would defeat the one check that stops a CLI verb shipping against a 404.

**Resolution is ordering, not a code change:**

1. merge ankraio/cluster#3173 (its CI is green bar the last couple of checks);
2. re-run this PR's CI — it will fetch the updated census and pass with no allowlist entry;
3. merge this.

Worth recording for anyone who hits this next: the **local pre-commit hook does not catch it.** `ANKRA_CLUSTER_ROUTES_JSON` is unset locally, so the test skips rather than fails, and `go test ./...` passed clean on my machine for every commit here. Locally you can reproduce CI with:

```
ANKRA_CLUSTER_ROUTES_JSON=../cluster/routes.json go test ./internal/client/ -run ClusterRoutes
```

PR #346, opened 2026-09-16
fix(backups): restore point sizes read "unknown" when nothing measured them (ankra-0xsdd.76)

CodeStaple:
## The defect

The fourth live verification pass (cluster `docs/runbooks/backups-verification-2026-09-16-fourth-pass.md`, item 5) sealed a restore point holding **48,217,836 bytes** of volume data — Velero `Completed`, 92/92 items, no warnings, no errors, the bytes confirmed in the bucket. The CLI rendered every size on it as `0 B`:

```
Size:          0 B               Volumes:        0 B
Assets:        1                 Databases:      0 B
  volume-notes-db │ volume │ velero │ crash │ notes │ db │ 0 B
```

`formatByteSize(0)` is `"0 B"`, and `RestorePoint.TotalBytes`, `RestorePointSizes.{Volumes,Databases}Bytes` and `RestorePointAsset.SizeBytes` were bare `int64` with no companion flag — so "nothing measured this" and "this is genuinely empty" printed identically. A person checking whether their backup worked was shown a number that says it did not.

## The fix

`describeRestorePointSize(sizeBytes, isKnown)` mirrors `restorePointOmissions` sitting thirty lines below it, which already solved exactly this for the not-carried column: the number when there is one, the word `unknown` when there is not. A **measured** zero still renders `0 B` — somebody established that, and it is a different fact from silence.

Applied at all six restore-point size sites in `cmd/cluster_stacks_restore_points.go` (the listing's SIZE column in both its per-stack and org-wide shapes, the detail's `Size`, `Volumes` and `Databases` lines, and the assets table) plus `describeLastRestorePoint` in `cmd/cluster_backups.go`.

The client structs carry the API's new flags — `total_bytes_known`, `size_bytes_known`, `sizes.known` — which is the `*Known bool` shape `DataAssetCountKnown`, `VaultsKnown`, `HasDatabaseKnown` and `ReadyVaultKnown` already use in this same package.

**They default to `false`, unlike `vaults_known`, and that is deliberate.** `vaults_known` defaults true so a server predating the field behaves as it always did; here "as it always did" *is* the defect. A server that does not send the field has never measured a restore point either, so unknown is the honest reading of its silence — and unknown is never the answer that misleads.

## Coverage

`restore-points get` also prints the manifest's new `coverage` block:

```
Coverage:
  namespace notes: restoring replaces everything in it, not only the 1 asset(s) named here (92 object(s) captured)
```

Velero's unit is the namespace, so a restore point naming one volume restores every object in that namespace. The pass found the sealed manifest did not say so — a reader deciding what a restore would overwrite had to find that out during the incident.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` green.
- `golangci-lint run` clean (`v2.12.2`, the version CI pins).
- `gofmt -l cmd internal` reports only `internal/client/baseurl.go` and `internal/client/helpers_test.go`, both untouched by this branch and unformatted on `master` already.
- New tests assert that an unmeasured restore point renders `unknown` and **never** `0 B` on the listing and on every line of the detail, and that `describeRestorePointSize(0, true)` still renders `0 B`.

Needs the matching `ankraio/cluster` PR deployed to serve the flags; until then every size reads `unknown`, which is the correct answer for every restore point sealed so far.

Bead: ankra-0xsdd-restore-point-sizes


CodeStaple:
Thanks — checked, and the two flags cannot disagree: the server derives both from the same value.

`ankraio/cluster` `go/internal/restorepointsapi/handlers.go`:

- `:464` — `TotalBytesKnown:    view.Manifest.Sizes.Known,` (this is `toListItem`, so it is on the listing as well as the detail)
- `:529` — `Known:          manifest.Sizes.Known,` (`toManifest`, which fills `sizes.known`)

Both read `manifest.Sizes.Known`, which is a single field on the sealed manifest, set once in `restorepoints.Seal` as `unmeasuredAssets == 0`. There is no per-kind measurement state on the server at all: `volumes_bytes` and `databases_bytes` are rolled up from the *same* per-asset sizes as `total_bytes` in the same loop, so "total measured but per-kind breakdown not" is not a state the seal can produce. A partial measurement makes all three unknown together, deliberately — a rollup over some of the assets, presented as a total, understates by an amount nobody can see.

I would rather not derive one from the other in the CLI: that would hide a genuine server bug behind a client-side guess, and rendering the mismatch is how we would find out. The invariant is now pinned on the server side instead — `ValidateManifest` refuses a manifest that carries a size while saying nothing measured it, so a sealed manifest cannot carry a number the flag disowns.

Both flags default to `false` on decode, so an older server that sends neither renders everything as `unknown` — consistent, and the honest answer for every restore point sealed so far.

PR #347, opened 2026-09-16
feat(ai): ankra ai remediation policy reads the auto-remediation policy (ankra-7vxh6)

Myothas:
Bead: ankra-7vxh6

## What & why

`ankra ai remediation policy` reads the organisation's AI auto-remediation
policy: the envelope every dispatched remediation run executes inside.

`GET /api/v1/org/ai-remediation/policy` is the only surface carrying
`autonomy_level`, the per-tool `tier_overrides` and the `cluster_allow_list`,
and nothing in the CLI exposed it. The P0 safety question in ankra-whdvi.6
(does `tier_overrides` name `apply_stack_changes` as `auto`, and what is the
cluster scope) needed a bearer token and a hand-written curl. `ankra ai status`
does not carry it, and `ankra ai autonomy` is the organisation-wide kill
switch, a different thing; its own help text pointed at the policy while
offering no way to read it, which this PR fixes too.

Read-only by design. The PUT is an organisation-admin, CSRF-protected browser
lane, and an API token that can widen autonomy is not a convenience worth
having.

**Naming.** `remediation` is a subgroup under `ai`, mirroring the platform's
own `/org/ai-remediation` route group, sitting beside `autonomy`, `lanes` and
`models`, and leaving room for the approvals inbox and the activity timeline
later. `policy` is the noun the platform uses.

### The two readings a safety review needs

- A `tier_overrides` entry of `auto` is labelled **"executes with NO approval
  card"**, and the count is repeated under the table. That is what the
  platform's tier ladder does with it (`scheduler/.../chatpipeline/remediation.go`,
  `ladderTier` then the `tier == TierApproval` park).
- A `cluster_allow_list` that was **never set** reads "every cluster (no allow
  list is set)". One that is **present and empty** reads "no cluster (the allow
  list is empty, so nothing matches it)". Those are genuinely different scopes
  on the platform side, so the CLI does not fold them together:
  `clusterInAllowList` in `scheduler/go/internal/maintloops/ai_remediation_dispatcher.go`
  admits every cluster for a null or absent list, and exactly the listed members
  otherwise, so an empty list admits none. The bead text says "empty means every
  cluster is in scope"; that holds for the unset case only.

### The no-row default

The platform answers an organisation with no policy row with its own default
**document** and an HTTP 200, never a 404, so "nothing is configured" arrives
looking exactly like a configured policy apart from a null `updated_at`.
Rendering that as somebody's decision would be a lie in the safety-critical
direction, so the command leads with "No auto-remediation policy is configured
for this organisation." and reports `Last saved: never`.

`-o json|yaml` (the repo's existing `registerStructuredOutputFlags`) carries
the same fact as a `configured` field, beside `cluster_scope`
(`all_clusters` / `listed_clusters` / `no_clusters`) and
`tools_without_approval`. The named scope is not decoration: the YAML encoder
renders a nil slice and an empty one identically as `[]`, so
`cluster_allow_list` alone cannot carry "never set" apart from "set to nothing"
outside JSON.

### Known platform gap, read this before merging

The bearer twin of the route does not exist yet. `airemediationapi` mounts the
policy on the session surface only (`/org/ai-remediation/policy`, `userAuth`),
and there is no `/api/v1` twin in `cliapi`. Verified against production on
2026-09-16 with no credentials:

```
/api/v1/org/ai-remediation/policy   -> 404   (no PAT route)
/api/v1/org/ai-settings/autonomy    -> 400   ("Bearer token not provided", the twin exists)
/org/ai-remediation/policy          -> 307   (session lane, redirect to login)
```

`routes.json` in the cluster census agrees: GET and PUT on
`/org/ai-remediation/policy`, nothing under `/api/v1`.

That cluster-side route is the other half of ankra-7vxh6 (the bead plans two
PRs) and overlaps ankra-us966.5's token-mount sweep. Until it lands:

- the command turns that 404 into "this platform does not serve the
  auto-remediation policy to API tokens: GET /api/v1/org/ai-remediation/policy
  is not registered", and exits 1 rather than 3, because exit 3 means "the
  targeted resource does not exist" and would read as "this organisation has no
  policy" - the opposite of what a 404 here proves;
- the path is recorded in `internal/client/cluster_routes_allowlist.json`, the
  ratchet `TestClusterRoutesAreRegistered` checks, beside the same-shaped
  `power-schedules` gap (ankra-htxvm.15), with the reason written into the
  test's header comment. It comes off that list when the cluster route exists,
  not before.

## Test plan

- `go build ./...` and the repo pre-commit gate (`go test ./...` +
  `golangci-lint run ./...`): all packages ok, **0 lint issues**.
- New tests: `cmd/ai_remediation_test.go` (7 cases: the no-row default is
  announced as the platform default and never as a saved policy; a saved policy
  is not called a default; the auto tier and the cluster scope are spelled out;
  an empty allow list stays apart from no allow list; `-o json` carries
  `configured`, `cluster_scope` and `tools_without_approval` for all three
  shapes; a 404 is reported as a missing route and does not exit 3; other
  failures relay with their own exit code) and
  `internal/client/airemediation_test.go` (4 cases, including that `null` and
  `[]` survive decoding as different values).
- `ANKRA_CLUSTER_ROUTES_JSON=<cluster origin/main routes.json> go test ./internal/client/ -run ClusterRoutes`:
  green with the recorded allowlist entry, and red without it, which is how the
  platform gap above was caught.
- End to end against a local fake serving the four response shapes, with the
  built binary: configured policy, no-row default, empty allow list, and the
  404. Rendered output for the first two:

```
$ ankra ai remediation policy          # organisation with no policy row
No auto-remediation policy is configured for this organisation.
What follows is the platform default, not a decision anybody made here.

Auto-remediation:   off, so no alert trigger opens an auto-remediation run
Autonomy level:     propose (every write waits for a human approval)
Cluster scope:      every cluster (no allow list is set)
Approvers:          none named (organisation admins decide)
Slack webhook:      none (approval cards have nowhere to go)
Max actions:        5 per incident
Cooldown:           30 minute(s)
Last saved:         never (no policy has been saved for this organisation)

Per-tool overrides: none, so every tool follows the autonomy level.

The organisation-wide stop is a different switch: ankra ai autonomy status.
```

```
$ ankra ai remediation policy          # configured policy
Auto-remediation:   on
Autonomy level:     auto (low-risk reversible writes execute unattended; everything else waits for approval)
Cluster scope:      every cluster (no allow list is set)
Approvers:          1 user(s)
  - 11111111-0000-4000-8000-000000000001
Slack webhook:      0b4f1b9e-0000-4000-8000-000000000001
Max actions:        7 per incident
Cooldown:           15 minute(s)
Last saved:         2026-09-01T10:30:00Z

Per-tool overrides:
╭─────────────────────┬──────────┬────────────────────────────────────╮
│ TOOL                │ TIER     │ WHAT IT MEANS                      │
├─────────────────────┼──────────┼────────────────────────────────────┤
│ apply_stack_changes │ auto     │ executes with NO approval card     │
│ delete_cluster      │ never    │ refused outright                   │
│ delete_pod          │ approval │ parks an approval card for a human │
╰─────────────────────┴──────────┴────────────────────────────────────╯
1 tool(s) run at the auto tier, so an auto-remediation run executes them with NO approval card.

The organisation-wide stop is a different switch: ankra ai autonomy status.
```

## Docs checklist

- [x] Command/flag changes — the CLI reference regenerates automatically on release (`tools/gendocs`); ensure `Short`/`Long`/`Example` on new commands are written for docs readers


Myothas:
The cluster-side half is now tracked on its own as ankra-7vxh6.1 (token-mount `GET /api/v1/org/ai-remediation/policy` plus the MCP read tool), filed separately so it survives this PR auto-closing ankra-7vxh6 on merge. The route-census allowlist entry and its paragraph in `cluster_routes_test.go` come off when that route exists; the ratchet fails on a listed path that resolves, so the cleanup is enforced.

Myothas:
Addressed in 40735ef: the missing-route rewrite now keys on the 404 status alone.

You were right to flag it even though the condition happened to hold. `getJSON` builds its non-200 error with `newUnexpectedResponseErrorWithMessage`, which leaves `Detail` empty for every status, and `NewUnexpectedResponseError` (the test constructor) goes through the same path, so the check was dead code rather than a live bug. But it was dead code pointed the wrong way: the moment `getJSON` learns to read a `detail` body, a platform 404 carrying one would have fallen through to exit 3, which is exactly the "this organisation has no policy" misreading the function exists to prevent.

`Detail` is the right key on a route that has a resource to miss. This one does not: it answers a default document for an organisation with no policy row, so a 404 means no route answered, whatever body came with it. There is now a test for a 404 that carries a detail.

Myothas:
**Held deliberately, not stalled.** Verified independently on cluster `origin/main`: `routes.json` carries `/org/ai-remediation/policy` (browser session lane) but no `/api/v1` twin, `cliapi` mounts only the `/api/v1/org/ai-remediation/approvals` prefix, and `openapi.json` has no `ai-remediation` path at all. So this command would return its "not registered" error for every user until the platform serves the policy to API tokens.

Merging it first would ship a command that cannot work. The token route is tracked as **ankra-7vxh6.1** and is being implemented now; this PR merges once that lands and deploys, and the `cluster_routes_allowlist.json` entry comes back out in the same pass (the ratchet fails on a listed path that resolves, so it cannot be forgotten).

Nothing to change here meanwhile: CI is green, the review verdict is "looks good", and the rendering keeps the three allow-list states apart, which is the part that matters.

PR #352, opened 2026-09-17
fix(pipeline): the CLI can see Ankra's platform builders (ankra-cssqw, PLA-868)

Myothas:
Bead: ankra-cssqw
Linear: PLA-868

A customer is about to run builds on Ankra's platform builders, and the CLI
cannot see that executor at all. A build the run's cluster cannot take is moved
to Ankra's own builders — a lane that opens no execution, and so has no live log
stream. Every CLI surface read that absence as "the step has not started".

### What changed

**`pipeline logs` tells a running platform build apart from a step that never
began.** The has-log-stream test reads `execution_id`/`execution_step_id`, which
a platform-builders step never gets, so a build in flight refused with `step
"build" has not started, so it has no log stream yet` and `--follow` announced
that it was waiting for a start that had already happened. The refusal now says
the build is running on Ankra's platform builders, that the lane has no live log
stream, and that this command prints the archived log once the step concludes;
`--follow` waits for that conclusion and prints the archive — the path that
already worked and that nothing pointed at. The step's `executor` is what tells
the two cases apart, so a step with no lane yet keeps today's refusal verbatim.
`platform` (gate, publish, approval) gets the same treatment for the same
reason.

**`pipeline get` grows an `EXECUTOR` column.** `PipelineStep.Executor` was on the
wire from the start and nothing printed it, so the one run detail a person reads
could not say a build had left their cluster. The column prints the platform's
own vocabulary (`in_cluster`, `platform_builders`, `platform`; `-` for a step
nothing has dispatched yet), and a short note under the table names the steps
Ankra's builders took and says where their logs are.

**`org ci-settings get` quotes the sentence a dark-grant build actually fails
with.** The note beside `platform_builders` quoted `build fallback is 'none'`.
Since cluster's `pipeline_step_dispatch`, that sentence is emitted only when the
setting really is none; a missing capability grant concludes `Ankra's
platform-operated build fallback is not enabled for this organisation yet`. The
note quotes that instead, so the text matches the step in front of the reader.

**`--build-fallback` help names both reasons a build leaves the cluster.** It
named only an agent that does not run pipeline steps. A cluster whose agent runs
them perfectly well still cannot build when its node runtime confines the
rootless builder (AppArmor/seccomp, as on k3s and Ubuntu nodes) — the case the
fallback most often carries.

**`org ci-settings set --help` says `platform_builds_enabled` cannot be set.** It
is Ankra's grant of the platform-builders capability, reported by `get` and
refused by the endpoint with a 422, so writing a `get` payload back wholesale
fails rather than half-applies.

### Tests

New tests pin each output string: the two `pipeline logs` answers for a
platform-builders step (with and without `--follow`, including that the
not-started refusal is unchanged for a step with no executor), the `EXECUTOR`
column and its note (and that a run which never left its cluster grows no note),
the quoted dark-grant sentence in both grant states, and both help texts.

No API or wire change; everything here reads fields the platform already sends.


Myothas:
Thanks — both findings are questions about code the diff does not show, and both cases are already covered on `master`. No change needed; here is the confirmation.

**[medium] `pipelineStepIsWaitingToStart` and a running platform build.** It does admit it. The function (unchanged by this PR, `cmd/pipeline_logs.go:1143-1152`) enumerates the states it waits in as `blocked`, `pending` **and** `running`:

```go
switch step.Status {
case pipelineStepStatusBlocked, pipelineStepStatusPending, pipelineStepStatusRunning:
	return true
default:
	return false
}
```

`running` is there for the in-cluster race it was written for — the row is claimed and its execution ids land a moment later — and a platform-builders step is `running` with ids that never land, so it takes the same branch. That is why this PR only adds a comment above it rather than touching the list: the states are right, and only the sentences were wrong. `TestPipelineLogsFollowWaitsForAPlatformBuildToConcludeAndPrintsItsArchive` drives exactly this path end to end (running platform-builders row → wait line → conclusion → archived log printed) and asserts no relay was opened.

**[low] Concluded platform-lane step without `--follow`.** The concluded check runs first, so that message is never reached. `runPipelineLogs` (`cmd/pipeline_logs.go:296-298`) routes on status before it looks at the stream:

```go
if step.Status == pipelineStepStatusConcluded {
	return runPipelineLogsFromArchive(command, selector, runID, step)
}
if !pipelineStepHasLogStream(step) {
	...
}
```

A concluded step therefore prints its archive immediately whether or not it ever had execution ids — `concludedStep()`, the fixture behind `TestPipelineLogsConcludedStepReadsArchivedStepLog`, carries no ids at all and reads its archived log fine. Reaching the "prints it once the step concludes" sentence requires a step that has *not* concluded, which is what it describes.

PR #355, opened 2026-09-18
ankra-4r75g.11/.12/.13: share the hidden-Unicode corpus, keep real joiners, strip structured output

Myothas:
Bead: ankra-4r75g.11
Bead: ankra-4r75g.12
Bead: ankra-4r75g.13

Stacked on #353 (base `fix/ankra-4r75g-cli-hidden-unicode`), which left the CLI's human-facing output clean but the contract unshared, the joiner rule too blunt, and machine output untouched. Three follow-ups, one per bead.

## Two things to push back on

**The `-o json|yaml` contract changes shape when a payload was hiding something.** An object payload now carries a `hidden_characters_removed` count, and a warning goes to stderr. Stripping silently was the alternative, and it is the thing machine consumers hate least in the moment and most later: a script handed quietly altered bytes has no way to learn the payload was hostile. That preference is my extrapolation from the human-facing lanes, not a stated requirement — **if you would rather keep the payload byte-exact and warn only on stderr, say so and I will drop the in-band field.** Note the strip only takes the generic path when something was actually removed, so output that was hiding nothing is byte-identical to before, field order included.

**The strip sits at `encodeStructured`, not in the three chat commands the bead named.** That function is the one seam all 25 structured callers pass through, so this also covers agent outcomes, remediation views, alert text and cluster names, and it needs no field-by-field walk of typed payloads (which is what the bead was wary of). The narrower fix would have left every other command exposed. Blast radius is the trade: every `-o json` command is on this path now.

## What each bead got

**.11 — the corpus.** `internal/hiddenunicode/testdata/corpus.json` is byte-for-byte the canonical file (md5 `34a3bb79c70f03bd61cefc3411c06176`), which `cluster enginekit/hiddenunicode` carries as the source of truth. `TestContractCorpus` asserts all 19 strip vectors; `TestVendoredCorpusMatchesCanonical` pins the checksum and the schema version, so editing the fixture to make the code pass fails the build — the failure worth stopping, because a test adjusted to fit the implementation protects nothing.

The 32-case table this repo added in #353 is retired. Every vector it held is covered by a canonical case except two, which moved to `TestRepoLocalExtraCases` with the reason they are local: **U+2068** (first strong isolate — the canonical bidi case covers U+202E/202C/2066/2069/200E/200F/061C but not this one) and **U+0001** (the canonical control cases use U+0007/001B/007F/000B/000C/0085/001C). Both belong upstream; they are here until the canonical file absorbs them.

Stated plainly, because it is the part that sounds stronger than it is: this checksum catches a **local** edit of the vendored copy. It cannot notice the canonical file moving ahead, because that file lives in another repository and cannot be read from a build here. The cross-repo gate is still open on ankra-4r75g.11.

**.13 — the joiner rule.** The strip removed every zero-width joiner and non-joiner except between emoji, so Persian and other scripts that write them inside ordinary words came back reshaped — a customer's own name, misspelled in what the CLI prints. `joinerIsLegitimate` now mirrors the cluster helper: kept between letters of a script that uses joiners, kept inside emoji sequences, still removed between Latin, Greek or Cyrillic letters, which is the case that hides text and makes `prod-1` and `prod<joiner>-1` render identically. `emojiSingles` gains the five code points the source of truth carries.

**.12 — structured output.** Covered above. Keys are stripped as well as values, since a hidden character in a key hides the key itself. An array or scalar payload cannot carry the marker without changing its shape, so it gets the stderr notice alone.

## The Indic correction, now settled upstream

This branch raised that the corpus's `known_limit` note was wrong: it said a joiner next to a combining mark is stripped, so "Indic conjuncts are reshaped by every implementation, including the source of truth". Cluster's `joinerIsLegitimate` skips `Mn`/`Mc` in its look-behind precisely because an Indic joiner follows the virama, so the real sequence survives. That was confirmed upstream, the note has been replaced with what the code does, and the canonical corpus now pins both orderings itself.

The corpus here is the frozen 21-vector revision, md5 `483986725b9a480a32eb8b890bdced3e`, and `TestVendoredCorpusMatchesCanonical` pins that hash. `TestJoinerRule` keeps both orderings pinned locally too, since they are the distinction most likely to rot:

- `devanagari zwj after the virama is kept` — consonant, virama, joiner, consonant: what real text produces.
- `devanagari zwj before the virama is stripped` — the same characters with the joiner on the wrong side, which is not a sequence real text produces.

## Gates

`go vet ./...` exit 0 · `go test -race -count=1 ./...` exit 0 (7 packages) · `golangci-lint` with `GOTOOLCHAIN=go1.26.6` **0 issues** · `gofmt` clean on every file this branch touches (`internal/client/baseurl.go` and `helpers_test.go` are unformatted on `master` and untouched here). Exit codes read directly, never through a pipe; lint and test runs serialised through the machine's lint mutex. Every changed file scanned free of literal invisible characters — writing `\uXXXX` through a file tool decodes it into the real character, which is the bug this package exists to stop.

CHANGELOG entries added per the repo's convention: one under Added for the structured-output change, one under Fixed for the joiner.


Myothas:
Both findings fixed in `839bbbc`, and both were worth catching.

**Numbers degrading in the stripped path.** Fixed with `json.Decoder.UseNumber()`, as suggested. The sharp edge is the one you named: the generic decode only happens when a payload *was* hostile, so without this the document whose numbers stopped being exact was exactly the document already under attack. `json.Number` is a distinct type, so the strip walk passes it through untouched and it marshals back as the literal it arrived as. Pinned by `TestStructuredOutputKeepsLargeNumbersExact` with 2^53+1 (the first integer a float64 cannot hold) and a nanosecond timestamp.

**Key collisions.** Fixed, and the randomness was worse than the overwrite: Go's map order is unspecified, so *which* value survived differed between runs of the same input. Keys are now visited in sorted order, the first keeps the cleaned name, later ones are dropped and counted, and the stderr notice says how many field names became identical. I went with "keep the first and say so" over documenting last-write-wins, because a document carrying both `cluster` and `clu<ZWSP>ster` is ambiguous before it reaches us, and reporting that is the point of this package. `TestStructuredOutputKeyCollisionIsDeterministicAndReported` runs the same input eight times and asserts identical output.

Also in this push: the canonical corpus was frozen at 21 vectors (md5 `483986725b9a480a32eb8b890bdced3e`) after this branch's Indic finding was confirmed upstream, so the vendored copy and the pinned checksum moved with it. Both new vectors pass against this package unchanged.

Gates on `839bbbc`: `go vet` 0 · `go test -race -count=1 ./...` 0 · `golangci-lint` 0 issues.

Myothas:
All three fixed in `ca9db02`. The YAML one was a genuine gap — my `UseNumber` fix covered only JSON, and the doc comment claimed more than the code did.

**YAML precision.** Took the first suggestion: the YAML path now walks a `yaml.Node` instead of decoding into `interface{}`, so every scalar keeps the literal it was written with and only `!!str` scalars are touched. Two things came free with it, which is why it beat narrowing the comment:

- **Key order survives.** The generic path sorted keys; the node walk keeps the document's own order.
- **Finding 2 disappears rather than being patched.** String keys are now cleaned wherever they appear, including inside a mapping whose other keys are not strings, because the walk sees nodes rather than a `map[interface{}]interface{}`. That branch is deleted — YAML no longer reaches the generic walk at all.

Pinned by `TestStructuredYAMLKeepsNumbersAndKeyOrder` (uint64 past int64's range, plus 2^53+1) and `TestStructuredYAMLCleansStringKeysBesideNonStringKeys`.

Worth recording: my first attempt at that test seeded its input with `yaml.Unmarshal` into `interface{}`, which degraded the numbers *before* the code under test saw them — the test failed and the failure was mine, not the fix's. The corrected test passes a typed struct, which is how the value actually arrives from a caller.

**Unconditional marker.** A payload that owns `hidden_characters_removed` now gets a suffixed marker rather than none, and its own field is left untouched. Agreed with the reasoning: stderr is not guaranteed to be captured, so the in-band signal should not be the thing that goes missing precisely when a document is adversarial. `TestStructuredMarkerIsUnconditional` asserts both halves.

Gates on `ca9db02`: `go vet` 0 · `go test -race -count=1 ./...` 0 (7 packages) · `golangci-lint` 0 issues.

Myothas:
Both fixed in `57fdc3a`. The medium one was wider than the diff could show, and you were right to ask me to confirm rather than assume.

**Direct callers got no notice.** I counted them: **55 call sites reach `encodeStructured` directly**, not through `renderStructured`. So this was not a theoretical path — it was most of them. An object payload still carried the in-band marker, but an array or scalar, which cannot carry it without changing the document's shape, was altered with no signal anywhere. That is the "strip and say so" contract degrading to "strip silently" across the majority of its call sites.

Took your second suggestion and moved the notice into `encodeStructuredCounting`, so every structured path emits it once. Callers holding a `*cobra.Command` pass its `ErrOrStderr` — this repo writes warnings that way throughout, and a package-level `os.Stderr` would have broken `SetErr` redirection — and the rest default to `os.Stderr`. `renderStructured` no longer writes it, and a test asserts the notice appears **exactly once** so the two paths cannot start duplicating it.

`TestStructuredNoticeReachesDirectCallers` covers object, array and scalar payloads through the direct path; `TestStructuredNoticeIsSilentForCleanPayloads` pins that a clean payload still says nothing.

**Fallback name could still collide.** Fixed: it now counts until it finds a free name rather than falling through to an untested one. The map is finite so it terminates. Agreed on the reasoning — a payload owning the marker plus a hundred suffixes is contrived, and contrived is what the adversarial path looks like; the previous code would have overwritten a field, which is the exact outcome that function exists to prevent. `TestMarkerFallbackNeverCollides` seeds 250 variants plus the old `_cli` fallback and asserts the chosen name is free.

Gates on `57fdc3a`: `go vet` 0 · `go test -race -count=1 ./...` 0 (7 packages) · `golangci-lint` 0 issues · gofmt clean.

PR #357, opened 2026-09-19
feat(pipeline,operations): validate a candidate at a ref, and cut the reconcile noise out of operations list (ankra-p20k5, PLA-863)

CodeStaple:
Two of the four CLI gaps Smartoptics raised on PLA-863 (Ankra #1185). The other two — `pipeline run --sha` taking another checkout's HEAD (#316) and `cluster addons upgrade --set` inlining a `from_file` values reference (#315) — shipped in v0.17.0.

## Ask 2: `pipeline validate` could not check a candidate

`validate` read the working tree or the stored definition, so finding out whether a change was valid meant merging it to the default branch and running it.

- `--ref <git-ref>` reads the definition as it stands on a git reference in the current checkout: `ankra pipeline validate --ref origin/my-branch --application my-app`. The reference is resolved locally (someone else's branch needs a `git fetch` first) and the path is read from the repository root, so it does not depend on which subdirectory the command was typed in.
- A reference that does not carry the file is an error, not a quiet fall-back to the stored definition — "ok" would answer a question the caller did not ask.
- `--spec-file` is the flag spelling of the file argument, matching `pipeline run --spec-file`. Naming the file both ways is a usage error (exit 2).
- `application pipeline validate` takes both flags, so the two surfaces do not drift.

## Ask 4: the operations list drowned by a reconcile loop

On so-upcloud-development 16 of the latest 50 executions were the same successful `Update addon defectdojo`, one every ~2 minutes, which pushed the failures that explained an outage off the first page.

- Consecutive identical **successful** executions fold into one row marked `(xN)` spanning the whole run's timestamps. Nothing that failed, and nothing whose name differs from the row beside it, is ever folded. `--no-collapse` lists every execution; `-o json|yaml` is unchanged.
- A folded run of five or more prints a line naming it as an add-on reapplying without a change in desired state rather than that many real changes — the third clause of the ask, on the CLI side. The defectdojo loop itself is ankra-7vojy.
- `--name` filters by add-on, manifest or stack name, case-insensitively, on the display name or the stored name. The executions API takes no name filter (verified against cluster `origin/main` `go/internal/execapi/executions.go`), so the listing walks its pages rather than filtering one of them — otherwise the filter would hide exactly what it is for. The walk is bounded at 20 pages and stops as soon as `--limit` is satisfied.
- `--limit` above 100 walks the same way instead of sending a `page_size` the API refuses (its max is 100), which was the "`--limit` tops out at 100" half of the complaint.

## Tests

`cmd/cluster_operation_test.go`: the fold, what must never be folded (failures, a different name beside it), the timestamp span, the name match, the loop notice, the page walk, the stop-at-limit, the single-call path when no walk is needed, and both flags driven through the command surface. `cmd/pipeline_test.go`: `--ref` reading the committed definition rather than the working tree, a ref without the file, `--spec-file`, and the file named twice.

`go build`, `go test ./...` and `golangci-lint run` are green (pre-commit hook ran both).

Bead: ankra-p20k5


CodeStaple:
Thanks, all three addressed in e66936b.

**Capped page walk reported as 'no executions matching'** — fixed. `fetch` now returns an `executionsPage` carrying `stoppedAtCap`, set only when the walk exhausted its 20-page budget with pages still unread. An empty capped listing says how far back it looked and what to narrow with (`--failed`/`--status`, or a shorter name) instead of claiming nothing matched; a capped listing that did find rows prints the same caveat under the table, and under `-o json|yaml` it goes to stderr so stdout stays parseable. Tests: `TestExecutionsQueryReportsACappedWalk`, `TestExecutionsQueryDoesNotReportACappedWalkWhenItReadEverything`, `TestPrintExecutionsReadCapNoticeOnlySpeaksForACappedWalk`.

**Collapse keys on DisplayName only** — fixed. Folding now goes through `isSameExecutionOperation`, which requires `DisplayName`, `Name` and `Type` to agree, so two operations sharing a label (or both carrying an empty one) stay separate rows. Test: `TestCollapseRepeatedExecutionsKeepsTwoOperationsSharingADisplayName`.

**Lexical timestamp comparison** — fixed. `chooseTimestamp` parses both values as RFC 3339 and compares instants, falling back to the text comparison only when one will not parse, so a mix of `Z` and `+02:00` cannot give a collapsed row the wrong window. Test: `TestCollapsedRunSpansAMixOfTimestampOffsets`.

On the `arguments[1:]` note in the summary: `application pipeline validate` is `Args: cobra.RangeArgs(1, 2)`, so there is always at least the application id before the slice. It is outside the diff, which is why it did not read as safe.

